PR #13, opened 2019-08-09
Add blog posts from ddl site

fennifith:
Moving the blog posts from [doubledot.dev](https://doubledot.dev/) to this repo. The site didn't build initially, until I realized my mistake and added the missing author data.


crutchcorn:
An interesting side effect of having this here is that "the site's first post" is no longer the first in the timeline, but oh well 🤷‍♂️

Issue #9, opened 2019-08-08
Add an about us page 


crutchcorn:
Designs provided by @tommyemo 

![about](https://user-images.githubusercontent.com/9100169/62716581-3ae8cd80-b9b7-11e9-9c30-4d601c6a7132.png)
![landing – mobile](https://user-images.githubusercontent.com/9100169/62716583-3ae8cd80-b9b7-11e9-943d-3bde47d8bafc.png)

adueppen:
I noticed that the authors page (as in **/authors**/crutchcorn) currently doesn't exist, but it seems like it would be pretty similar to the contributors section for the about page. I'm not quite sure how we would want to structure it but it's something to keep in mind.

crutchcorn:
I mean, we can just add an HTML redirect of `/authors` to `/aboutus#authors`, I think that would suffice

adueppen:
Yeah I was kind of thinking about that but you have a better idea of how you want to organize the site anyway so I was going to leave it up to you.

PR #15, opened 2019-08-10
About us page, user roles, rename `authors` to `unicorns`

crutchcorn:
This PR implements #9, trying to match design as much as possible

It also adds and changes a few things pertaining to the about us page and contributors in general:
- [x] Renames `authors` to `unicorns`
This drove some discussion within the team (on our Discord). Originally the idea of `users` was suggested as an alternative to `authors`, as our long-term goals include the ability to have accounts to track user's progress (as we're wanting to shift to a more structured educational platform in the future). However, there was decent conversation about calling folks who learn and contribute to our platform "users" being inpersonal. We then decided to call them `unicorns` to fit our overall branding. 
As a result, a 302 redirect will need to be added to our apache config (will come in a future PR from the @unicorn-utterances/devops team when we're ready to deploy) from `/authors/crutchcorn` (the only current route under the old URL structure) to `/unicorns/crutchcorn`
- [x] Added the ability to add roles
Per the previous point, we wanted a way to clearly communicate the achievements of our contributors/members that may not be authors to the site. We now have the ability to list a role or contribution method on a per-unicorn basis


crutchcorn:
It's worth mentioning that @adueppen had reminded that a way to redirect from `/unicorns` to the list of contributors on this about page would be advantageous, so the @unicorn-utterances/devops team will also need to include a 302 from `/unicorns` to `/about#contributors` (as we have ensured that the anchor point there works)

content/site/about-us.md
```diff
@@ -0,0 +1,18 @@
...
+
+> Our goal is to provide the best resources possible to learn any computer science related topic
+
+Whether it's from how memory is allocated in assembly or how to do complex CSS animations, to what a For loop is, we want to provide resources that are welcoming, supportive, and informational.
+
+Our content will have a wide range of subject matters, difficulty curves, and (we hope) perspectives. We want to make resources for the seasoned full-time developer just as we would a newcomer hobbyist and everyone between the two.
+
+ We know this is a lofty goal, though, and we don't want to do it alone. If you're interested in helping out, open a pull request and help us reach others by providing translations, help the code maintenance on this site, write a post for us, or jump in on discussions and help other readers.
+ 
+ If you’d like to get in touch with us, our GitHub is probably the best way to do so.
```
fennifith: You might want to actually add email / contact links somewhere on the page if you're mentioning it here, as I don't think there are any other references to the GitHub org other than the "view on GH" link at the bottom of the articles.

Maybe a small footer on each page linking to the about section, github, email, social media, etc. would do the trick.

crutchcorn: I would love to add social media and email but we currently don't have either. I will add actual links to what I was talking about though

Issue #4, opened 2019-06-30
Add dark mode


crutchcorn:
In the `src/assets/icons` directory, we have a `dark` and `light` icons that were designed for the site. This feature was cut due to time, but having in toggle in the top right of the header (#3) would make for a better reading experience for users that prefer a dark mode

Because we're using CSS variables, this should be relatively straightforward

crutchcorn:
Attaching designs from the one-and-only amazing @tommyemo 

![about](https://user-images.githubusercontent.com/9100169/61683264-caa92f00-acc9-11e9-8b36-0781f379f05f.png)
![author – searched – expands-1](https://user-images.githubusercontent.com/9100169/61683265-caa92f00-acc9-11e9-8791-dcb54c65083e.png)
![author – searched – expands](https://user-images.githubusercontent.com/9100169/61683266-caa92f00-acc9-11e9-9a01-7ad7b1682d83.png)
![author](https://user-images.githubusercontent.com/9100169/61683267-cb41c580-acc9-11e9-85ec-88e47ac6ed3a.png)
![blog detail view – bottom](https://user-images.githubusercontent.com/9100169/61683268-cb41c580-acc9-11e9-8d4a-3d419fca3aed.png)
![blog detail view – overflow](https://user-images.githubusercontent.com/9100169/61683269-cb41c580-acc9-11e9-9773-1548b6e03070.png)
![blog detail view-3](https://user-images.githubusercontent.com/9100169/61683270-cb41c580-acc9-11e9-87ca-560baf6c593e.png)
![landing – _edited for length_](https://user-images.githubusercontent.com/9100169/61683271-cb41c580-acc9-11e9-8ea6-f83b86f9f297.png)
![landing – scrolled](https://user-images.githubusercontent.com/9100169/61683272-cbda5c00-acc9-11e9-883a-b87bbe34f4a1.png)
![landing – searched + filtered](https://user-images.githubusercontent.com/9100169/61683273-cbda5c00-acc9-11e9-9077-f1f6b7739d8c.png)

tommyemo:
💜✨💥

adueppen:
Something that may require changing with a dark mode is the meta theme-color property. At the moment, it's #127db3, the background color of the logo. This has decent contrast but would look a bit odd in a dark mode.

crutchcorn:
[...]

PR #17, opened 2019-08-10
Dark mode

adueppen:
At the moment it's not a toggle button and the code is a bit messy. The color change doesn't seem to apply to SVG icons at the moment and I'm not quite sure why, but everything else seems to work OK enough.


adueppen:
(comment mentioning #4 for link)

adueppen:
The dark mode *should* be done now but there are some remaining contrast issues with code blocks in dark mode. At the moment I'm using the colors from the sulphurpool-dark color scheme so if we want to change it we might want to consider a different color scheme entirely.

crutchcorn:
Work still needed:

- [ ] Investigate changing some of the CSS with Disqus (this might be possible)
- [ ] Replace the Prism color vars with CSS variables 
- [ ] Add dark mode values to the prism color vars
- [ ] Change the styling of inline code (I believe this is still prism)

adueppen:
OK so the dark code styling should be complete now, I have an idea for how to make the Disqus thing work but I think I might be making a mistake with namespacing.

crutchcorn:
Thank you so much for your patience and consistent dedication to polish for the feature @adueppen

PR #27, opened 2019-08-12
Multiple post authors

adueppen:
We don't currently have any multi-author articles so there is a test article included in this branch which we'll have to make sure to remove before merging.


gatsby-config.js
```diff
@@ -133,7 +133,7 @@ module.exports = {
                   url: nodeUrl,
                   guid: nodeUrl,
                   custom_elements: [
-                    {"dc:creator": frontmatter.author.name },
+                    {"dc:creator": frontmatter.authors.name }, //changed
```
crutchcorn: This will not fail on develop, but will fail on build. We need to make sure to loop back to fix issues with RSS and ensure it handles single and multi-authors properly. We do currently have users that use our RSS feed so we should keep consistent behavior

gatsby-config.js
```diff
@@ -224,7 +224,7 @@ module.exports = {
             title: node => node.frontmatter.title,
             content: node => node.rawMarkdownBody,
             slug: node => node.fields.slug,
-            author: node => node.frontmatter.author.name,
+            authors: node => node.frontmatter.authors.name, //changed
```
crutchcorn: Likewise, this will also need to be updated

src/templates/blog-profile.js
```diff
@@ -64,7 +64,7 @@ export const pageQuery = graphql`
     }
     allMarkdownRemark(
       filter: {
-        frontmatter: {author: {id: {eq:  $slug}}},
+        frontmatter: {authors: {elemMatch: {id: {eq:  $slug}}}},
```
crutchcorn: elemMatch? I've not heard of that yet - I'll have to check it out :)

src/templates/blog-post.js
```diff
@@ -30,7 +30,7 @@ const BlogPostTemplate = (props) => {
       <SEO
         title={post.frontmatter.title}
         description={post.frontmatter.description || post.excerpt}
-        unicornData={post.frontmatter.author}
+        unicornData={post.frontmatter.authors} //might need to do CSV list here
```
crutchcorn: We don't want to do this. If we did, we would lose lots of what makes our GraphQL server so nice to use

src/templates/blog-profile.js
adueppen: is doesn't seem to be a thing, but I think elemMatch is what you're looking for.

src/templates/blog-post.js
adueppen: I guess I'm just not quite sure how we could otherwise present the list of authors.

src/components/post-view/post-metadata/post-metadata.js
```diff
@@ -4,18 +4,22 @@ import styles from "./post-metadata.module.scss"
...
       <div className={styles.textDiv}>
-        <Link to={`/unicorns/${author.id}`} ref={authorLinkRef} className={styles.authorLink}>
-          <h2 className={styles.authorName}>{author.name}</h2>
-        </Link>
+        <h2 className={styles.authorLink}>
+          {authors.map(author => {
+            return <Link to={`/unicorns/${author.id}`} ref={authorLinkRef} className={styles.authorName}>{author.name}</Link>
+          }).reduce((prev, curr) => {
+            return [prev, ", ", curr]
+          })} {/*it works but it's not pretty, also messes up styling a bit*/}
```
crutchcorn: Let's talk more on Discord

crutchcorn:
image

Still a WIP, but this is the current styling applied for multi authors in the list view I worked on last night. @tommyemo would love your thoughts on this? 

Here it is in the grid:

image

adueppen:
I think everything here is mostly good, although I did notice that the profile pics can get a bit too close to the post metadata on small screen widths. I remember talk of aligning them vertically at small widths so we might want to consider trying to do that.

src/components/post-card/post-card.js
```diff
@@ -73,12 +73,13 @@ export const PostCard = ({ title, authors, published, tags, excerpt, description
...
-          <span onClick={getAuthorLinkHandler(0)}>by&nbsp;{authors[0].name}</span>
-          {authors.slice(1).map((author, i) =>
-            <span key={author.id} onClick={getAuthorLinkHandler(i + 1)}>, {author.name}</span>
-          )}
+        <p className={cardStyles.authorName}>
+          <span>by&nbsp;</span> {/*the multiple spans can probably be cleaned up a bit*/}
+          <span onClick={getAuthorLinkHandler(0)} className={cardStyles.authorLink}>{authors[0].name}</span>
+          {authors.slice(1).map((author, i) => [
+            <span>, </span>,
+            <span key={author.id} onClick={getAuthorLinkHandler(i + 1)} className={cardStyles.authorLink}>{author.name}</span>
```
adueppen: As my comment suggests I know these `span`s can probably be cleaned up a bit with some ternary operators or something, I just didn't want to make it too complex.

crutchcorn: Actually, I prefer having this split up. The code is more readable. I know I'm guilty of "clever" code in the codebase (which I'm trying to cleanup), but readability should come first

crutchcorn:
The code should now be SIGNIFICANTLY cleaner, handle various sizing better, and is overall more modular and better. Please re-review @adueppen

adueppen:
Looks like all the bugs have been worked out, so I think we're ready to merge!

PR #29, opened 2019-08-18
Change about page link to a <Link> component

adueppen:
This allows it to load in place instead of fully reloading the page.


crutchcorn:
Great catch, let's merge!

PR #33, opened 2019-08-22
Fix posts' URL redirect

crutchcorn:


crutchcorn:
This is not needed in master before it shows in prod. This file is not actually used in our live server config (for security reasons), but is meant to be tracked alongside (for redundancy/transparency)

PR #39, opened 2019-08-23
Change social icons to components instead of functions

adueppen:
This makes sure the SVG attributes remain intact, critically including the class which is needed for some CSS rules.


crutchcorn:
I'm still not entirely sure that this change preserves the class (or why it would if it does), but I'm in favor of this change otherwise anyway

PR #40, opened 2019-08-23
Make sure focus ring has good contrast

adueppen:
For some reason Firefox has a somewhat different thing going on with the focus ring. This solution works fine on both Linux and Windows so far, which matters due to the fact that Firefox attempts to use native OS UI styles for things such as buttons.

### DIFF
--- a/src/global.scss
+++ b/src/global.scss
@@ -32,6 +32,16 @@
   }
 }
 
+:focus {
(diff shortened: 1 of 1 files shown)


crutchcorn:
This works great. Love the insight of Firefox's workings and focus on accessability on display here!

PR #44, opened 2019-08-24
WIP: Page transition

crutchcorn:
This PR adds in a page transition effect between all pages. 

It uses `gatsby-transition-link` and `react-posed`. While there is an `AniLink` component, it uses `gsap` and we're already using `react-posed`, so I wanted to avoid dependency duplication

Some work needs to be done:

- [ ] Code cleanup
- [ ] Fix on mobile when the user is scrolled down (it jumps up)
- [ ] Have back button fade out when disappears
- [ ] Fix the animation when opening an author page from the homepage


adueppen:
This is just my opinion, but it would be nice to have some way of turning the transitions on or off. I generally find them nice to have when viewing a site on mobile, but I don't like them at all on desktop sites.

evelynhathaway:
While it wouldn't solve @adueppen's issue, we should add support for [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) at the same time as creating a switch.

crutchcorn:
Right, that was the plan @evelynhathaway 

We should make a new issue for it

crutchcorn:
I'm gonna close this PR and we can investigate re-working this stuff in the future. Right now, it's simply too outdated and broken to consider this specific PR

PR #47, opened 2019-08-30
Added emotes and logo SVGs

crutchcorn:
This PR adds what is quite possible, the most cute emotes ever made by our logo designer @zavukodlak

To see a preview of these emotes, visit this page (part of this commit):
https://github.com/unicorn-utterances/unicorn-utterances/blob/d498dbc3035bc0a0766ddf67163c29d2e501362e/content/assets/branding/emotes/README.md

Please not that this PR also adds the logo SVGs to keep all of our material (including our branding work). However, these assets are under [the CC by NC ND license](https://creativecommons.org/licenses/by-nc-nd/4.0/). This means they cannot be sold, remixed, and must be attributed


adueppen:
The one thing we might want to change first is to optimize at least a few of the PNGs, the emote-parts color files are a bit big at the moment so I feel like they should either be optimized or perhaps have a slight grain effect.

crutchcorn:
While I generally agree - I'd like to keep these images as-high-quality as possible and then use our code-gen tools to downsize these. They won't be hosted on our server during build

PR #57, opened 2019-09-06
Initial mobile styling

crutchcorn:
This PR does some initial work to improve mobile styling and consolidating of the media queries used to make mobile styling work

Something that @tommyemo pointed out is that it does not match his mobile mockup for the posts properly - this is something we'll want to address in a (close-to-now) PR with him (which I intend to be doing later in the next week)

### DIFF
--- a/src/_vars.scss
+++ b/src/_vars.scss
@@ -2,7 +2,10 @@ $baseUnit: 8;
 
 $rootFontSize: #{$baseUnit * 2}px;
 $startMediumScreenSize: 768px;
-$endMobileScreenSize: $startMediumScreenSize - 1px;
+$endSmallScreenSize: $startMediumScreenSize - 1px;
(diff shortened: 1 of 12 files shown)


src/_vars.scss
```diff
@@ -2,7 +2,10 @@ $baseUnit: 8;
 
 $rootFontSize: #{$baseUnit * 2}px;
 $startMediumScreenSize: 768px;
-$endMobileScreenSize: $startMediumScreenSize - 1px;
+$endSmallScreenSize: $startMediumScreenSize - 1px;
+$startSmallScreenSize: 440px;
+$startSuperSmallScreenSize: 348px;
+$endSuperSmallScreenSize: $startSmallScreenSize - 1px;
```
crutchcorn: Can we move this line one higher? I was a bit confused at first where this was 😅

PR #58, opened 2019-09-07
Fix: Add extra newline after inline iframes in Angular Templates Post

evelynhathaway:
Having only one newline causes the markdown on the next line to not render correctly.

Replace: `/(?<=</iframe>)\n(?!\n)/` with `\n\n`


crutchcorn:
Blech, I tried doing this many times and yet somehow this STILL persisted. 

Thanks for the fix!

Issue #45, opened 2019-08-25
Back button should take user to previous page instead of homepage in some situations


adueppen:
Currently the back button is hardcoded to route the user back to the homepage which definitely makes sense for things like blog posts, but when going from homepage --> about page/blog post --> unicorn page, the back button there should return the user to the previous page they were on instead of the homepage.

crutchcorn:
I admittedly thought we had a story for this. Unfortunately, that's a bit complicated. 

We'd need this to be merged first: 
https://github.com/reach/router/pull/197

Otherwise, trying to add `location.back` would hard reload the last page which would break a lot of transition functionality/etc

crutchcorn:
I found an issue (now that reach/router#197 was merged) where `navigation` from `gatsby` doesn't allow -1 to be passed (the back command) due to it's query parsing. I am trying to talk to the Gatsby team about opening a PR myself to fix this issue, but have gotten stuck on a technical issue and am awaiting a reply from them. More as this develops

crutchcorn:
https://github.com/gatsbyjs/gatsby/pull/19044 we're now waiting on this

PR #61, opened 2019-09-12
A11y fixes for filter, subheader, profile social links, and card list, etc

crutchcorn:
This PR provides some fixes for A11Y on the site that does the following:

- Focuses the button on the listbox after escape is pressed
- Adds missing aria attributes for the filter listbox
- Change subheader on profile and about page from h2 to `aside` (which fits MUCH better for the UX of the page)
- Move profile social links to a `ul`
- Move the cards list to a `ul`

### DIFF
--- a/src/components/filter-search-bar/filter-listbox/filter-listbox.js
+++ b/src/components/filter-search-bar/filter-listbox/filter-listbox.js
@@ -124,6 +124,17 @@ export const FilterListbox = ({ tags = [], className, onFilter }) => {
   const windowSize = useWindowSize(150)
 
   /**
+   * Effects
+   */
+  useEffect(() => {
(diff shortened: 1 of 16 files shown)


adueppen:
OK so this seems to work well for the most part but I did notice that:
- The CI tag should be capitalized (otherwise screen readers seem to pronounce it like "psy")
- This will have to be changed after #45 is fixed but the back button should probably be labelled as "Go back to homepage" instead of just "Go back".
- The "no results found" text is in the :after pseudo-elements so `aria-live` needs to be used to announce it.
- Post title and info is aligned to the left.

crutchcorn:
The other issues should be defered to another PR, but should be worked on. Thanks for the review and the call today!

Issue #28, opened 2019-08-13
Filter button is deceptively small on mobile


adueppen:
At small screen widths, the filter button appears to take up the entire width of the screen, but it can only be activated by clicking/tapping on the icon and/or text. The search bar similarly expands to the width of the screen, but it does not have this issue. ![Screenshot](https://i.imgur.com/QSPZve6.png)

adueppen:
Closed by #76

PR #76, opened 2019-09-21
Fixed some spacing issues (as well as #28) with filter/search

adueppen:
Everything seems to work fine but I feel like there might still be some issues lurking. Hopefully this can work as a short-term fix.


crutchcorn:
I'm amazed you were able to edit this code at all - kudos. Seems to work well - merging. I'll be working more on refactoring this code today just as a heads up

PR #81, opened 2019-09-22
`useSelect`, `filterListbox` refactor

crutchcorn:
This is the first half of work for #2. This is a massive code cleanup for `useSelect`, but I believe this functions as it did before. This should all probably be broken out to a library of it's own and documented and tested better


crutchcorn:
Two issues present in this refactor:

- [ ] Pressing the button when open does not close it
- [ ] On initial page load, the arrow keys do not work (but manually placing focus does)

crutchcorn:
It seems the only issue left is the duplicate selection for the button on FF only. I suspect this is due to differences in how the `click` event is handled from Chrome, selecting an underlying element that the container el does not `contain`. 

After I get that solved, I'll be writing tests for all of the hooks

crutchcorn:
It seems there's some bugs pertaining to the styling of the filter listbox. However, it seems these bugs are present in the current `integration`. Let's fix those AFTER merging in this PR, since this PR will make that significantly easier

crutchcorn:
Okay, so I'm left unclear why FF still has the issue where if it's open and pressing it again, it will open it again, but I'm okay with leaving that behavior for a future PR to address. This PR has now a MASSIVE scope and I'd love to get it moved inwards before it grows too much more

Issue #82, opened 2019-09-22
Full results of search do not match


adueppen:
Searching for "freenod", "free", etc comes up with the freenode IRC posts, but searching for "freenode" comes up with no results. Similarly, "introduct" comes up with 2 articles with "introduction" in their names, but as soon as any of the last 3 letters are added it finds nothing. This will need some more testing to figure out the root cause.

crutchcorn:
This also breaks filtering - where we're using `lunr` to use the filter

PR #83, opened 2019-09-22
Lazy fix of full results in Lunr

crutchcorn:
This PR is a temporary fix. It adds too many results - I'm finding that I don't know Lunr well enough to fix the issues with the site long-term


crutchcorn:
Fixes: #82

PR #86, opened 2019-09-27
Added typescript generics post

crutchcorn:
Any feedback (tone, grammar, spelling, etc) hugely welcome. I'm excited to go through the first open dialog regarding a PR for a post! :D

### DIFF
--- a/content/blog/typescript-type-generics/index.md
+++ b/content/blog/typescript-type-generics/index.md
@@ -135,7 +135,7 @@ const returnedObject = returnSelf({objProperty: 12});
 // This will fail, as expected
 returnedObject.test();
 // This will exist, and TS will know it as a number
-console.log(returnedObject.objProperty);
+returnedObject.objProperty;
 ```
 
 > Author's note:
--- a/content/blog/typescript-type-generics/index.md
+++ b/content/blog/typescript-type-generics/index.md
@@ -105,7 +105,7 @@ function returnSelf(returnProp: any): any {
 const returnedObject = returnSelf({objProperty: 12}); // This now works! 🎉
 
 returnedObject.test(); // This will not return an error but should 🙁
-returnedObject.objProperty; // This will also (correctly) not throw an error, but TS will not know it's a number ☹️
+returnedObject.objProperty; // This will also (correctly) not throw an error, but TS won't know it's a number ☹️
 ```
 
 # The Real Solution {#generics-intro}
@@ -182,7 +182,7 @@ async function logTheValue(item) {
 }
 ```
 
-If we wanted to type the `logTheValue` function, we'd want to make sure to use a type generic for the input parameter `item`. By doing so, we could use that same geneeric for the return prop of `loggedValue` to ensure they both have the same typing. To do this, we could do so inline:
+If we wanted to type the `logTheValue` function, we'd want to make sure to use a type generic for the input parameter `item`. By doing so, we could use that same generic for the return prop of `loggedValue` to ensure they both have the same typing. To do this, we could do so inline:
 
 ```typescript
 // Because this is an `async` function, we want to wrap the returned type value in a Promise
@@ -191,7 +191,7 @@ async function logTheValue<ItemT>(item: ItemT): Promise<{loggedValue: string, or
 }
 ```
 
-Alternatively, we could utilize another feature of generics: the ability to pass the type value of the generic manually, and make an interface with a generic and do so there:
+Alternatively, we could utilize another feature of generics: the ability to pass the type value of the generic manually by making an interface with a generic and doing so there:
 
 ```typescript
 interface LogTheValueReturnType<originalT> {
@@ -252,8 +252,8 @@ interface ImageType<DataType> {
 
 interface ImageConvertMethods<DataType> {
   // This is the typing of a method. It will take a prop of the generic type and return the generic type
-  toPNG(data: DataType) => DataType;
-  toJPG(data: DataType) => DataType
+  toPNG: (data: DataType) => DataType;
+  toJPG: (data: DataType) => DataType;
 }
 
 
@@ -296,8 +296,8 @@ Unfortunately, there's a problem with the above code: we don't know what type `D
 Let's fix that typing:
 
 ```typescript
-function toPNG(data: DataType extends (string | Array<number> | Buffer)): DataType {
-// ...
+function toPNG<DataType extends (string | Array<number> | Buffer)>(data: DataType): DataType {
+	// ...
 }
 ```
 
@@ -313,7 +313,7 @@ interface TimestampReturn<T> {
 	isFuture: boolean;
 	obj: T
 }
-const checkTimeStamp = (obj: T extends {time: Date}): TimestampReturn<T> => {
+const checkTimeStamp = <T extends {time: Date}>(obj: T): TimestampReturn<T> => {
 	let returnVal: TimestampReturn<T> = {
 		isPast: false,
 		isFuture: false,
--- a/content/blog/typescript-type-generics/index.md
+++ b/content/blog/typescript-type-generics/index.md
@@ -14,7 +14,7 @@ While working in various projects, you may come across a weird looking syntax in
 
 _Type generics are a way to handle abstract types in your function._ **They act as a variable for types in that they contain information about the way your types will function.** They're very powerful in their right, and their usage is not just restricted to TypeScript. You'll see many of these concepts applied under very simiar terminology in various language. Enough on that, however, let's dive into how to use them! 🏊‍
 
-# The Problem #{generic-usecase-setup}
+# The Problem {#generic-usecase-setup}
 
 Type generics — on the highest level — _allow you to accept abitrary data in place of strict typing in order to broaden up a type's scope_.
 
@@ -33,7 +33,7 @@ returnProp(4); // ❌ This would fail as `4` is not a string
 
 In this case, we want to make sure that every possible input type is available for the prop type. Let's take a look at a few potential solutions with their various pros and cons and see if we can find a solution that fits our requirements for typing a function like this.
 
-## Potential Solution 1: Unions #{generic-usecase-setup-union-solution}
+## Potential Solution 1: Unions {#generic-usecase-setup-union-solution}
 
 One potential solution to this problem might be TypeScript Unions. _TypeScript unions allow us to define an `or` condition of sorts for our types_. As we're wanting to allow for various types for inputs and outputs, perhaps that might help us here!
 
@@ -58,7 +58,7 @@ const newNumber = shouldBeNumber + 4;
 
 The reason the opperation `shouldBeNumber + 4` yeilds this error is because it doesn't know which to give you given your input, given that you've explicitly typed your output as also being a union (meaning that it could be a string OR a number returned — one of them supporting numeric operations one of them not).
 
-### Potential Solutions Disclaimer #{silly-examples-disclaimer}
+### Potential Solutions Disclaimer {#silly-examples-disclaimer}
 
 > Author's note:
 >
@@ -66,7 +66,7 @@ The reason the opperation `shouldBeNumber + 4` yeilds this error is because it d
 >
 > That said, we're trying to build on concepts so we're trying to provide some examples of where this might be used and what it does. There are also instances, such as type definition files, where this inference might not be available to an author of typings, and that's ignoring other limitations with this method we'll see later
 
-## Potential Solution 2: Function Overloading #{generic-usecase-setup-overloading-solution}
+## Potential Solution 2: Function Overloading {#generic-usecase-setup-overloading-solution}
 
 In order to get around the issues with explicitly returning a union, you COULD utilize function overloading to provide the proper return typings:
 
@@ -91,7 +91,7 @@ returnProp({}) // Argument of type '{}' is not assignable to parameter of type '
 
 This may seem obvious from the typings, but _we ideally want `returnProp` to accept ANY type because **we aren't using any opperations that require knowing the typing**._ (no addition or subtraction to require a number, no string concatenation that might restrict an object from being passed)
 
-## Potential Solution 3: Any #{generic-usecase-setup-any-solution}
+## Potential Solution 3: Any {#generic-usecase-setup-any-solution}
 
 Of course, we could use the `any` in order to force any input and return type. (Goodness knows I've had my fair share of typing frustrations that ended with a few `any`s in my codebase!)
 
@@ -108,7 +108,7 @@ returnedObject.test(); // This will not return an error but should 🙁
 returnedObject.objProperty; // This will also (correctly) not throw an error, but TS will not know it's a number ☹️
 ```
 
-# The Real Solution #{generics-intro}
+# The Real Solution {#generics-intro}
 
 So what's the answer? How are we able to get both preserved type data on the input as well as the output??
 
@@ -144,7 +144,7 @@ returnedObject.objProperty;
 >
 > Remember, type variables are like other variables in that you need to maintain them and understand what they're doing in your code
 
-# Okay, but Why? #{logger-example}
+# Okay, but Why? {#logger-example}
 
 Why might we want to do this? [Returning an item as itself in a identity function](#generic-usecase-setup) is cool, but not very useful in it's current state. That said, there ARE many many uses for generics in real-world codebases.
 
@@ -214,7 +214,7 @@ An example of this would be a syntax like this:
 logTheValue<number>(3);
 ```
 
-# Non-Function Generics #{non-function-generics}
+# Non-Function Generics {#non-function-generics}
 
 As you saw before with the `LogTheValueReturnType` interface — functions aren't the only ones with generics. Additional to using them within functions and interfaces,  you can also use them in classes. 
 
@@ -258,7 +258,7 @@ interface ImageConvertMethods<DataType> {
 type ImageTypeWithConvertMethods<DataType> = ImageType<DataType> & ImageConvertMethods<DataType>
 ```
 
-# Okay, but why-_er_? #{polymorphic-functions}
+# Okay, but why-_er_? {#polymorphic-functions}
 
 Well, "I want to see your credentials" person you, type generics enable us to do things like type **polymorphic functions**. _Polymorphic functions are functions that can accept a myriad of different types and handle them differently._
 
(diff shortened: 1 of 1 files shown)


content/blog/typescript-type-generics/index.md
```diff
@@ -48,54 +48,54 @@ returnProp('Test'); // ✅ This is fine
...
 ```typescript
 function returnProp(returnProp: number): number;
 function returnProp(returnProp: string): string;
-// While this seems duplicative, TS requires it
+// While this seems repetitive, TS requires it.
 // Otherwise, it will complain:
 // This overload signature is not compatible with its implementation signature.
 function returnProp(returnProp: string | number): string | number {
-	return a;
+	return returnProp;
```
crutchcorn: Great catch :D

content/blog/typescript-type-generics/index.md
```diff
@@ -105,16 +105,16 @@ function returnSelf(returnProp: any): any {
 const returnedObject = returnSelf({objProperty: 12}); // This now works! 🎉
 
 returnedObject.test(); // This will not return an error but should 🙁
-returnedObject.objProperty; // This will also (correctly) not throw an error, but TS will not know it's a number ☹️
+console.log(returnedObject.objProperty); // This will also (correctly) not throw an error, but TS will not know it's a number 🙁
```
crutchcorn: I'd like to revert that. I know TS is complaining but that's also sorta the point 😅Having the `returnedObj` on different starting indexes makes this harder to scan

```suggestion
returnedObject.objProperty; // This will also (correctly) not throw an error, but TS will not know it's a number ☹️
```

I also intentionally made the emoji more sad 😜

content/blog/typescript-type-generics/index.md
```diff
@@ -124,29 +124,29 @@ function returnSelf<T>(returnProp: T): T {
...
 // Likewise, this object is now returned as if it was just placed on the const
 const returnedObject = returnSelf({objProperty: 12});
 
-// So this will fail, as expected
+// This will fail, as expected
 returnedObject.test();
-// And this will exist and TS will know it as a number
-returnedObject.objProperty;
+// This will exist, and TS will know it as a number
+console.log(returnedObject.objProperty);
```
crutchcorn:
```suggestion
returnedObject.objProperty;
```

content/blog/typescript-type-generics/index.md
```diff
@@ -182,15 +182,15 @@ async function logTheValue(item) {
 }
 ```
 
-If we wanted to type the `logTheValue` function, we'd want to make sure to use a generic for the `loggedValue` was using the same generic type as `item`. We could do so inline:
+If we wanted to add typings for the `logTheValue` function, we'd want to make sure to use a generic for the `loggedValue` <!--Looks like some sentence revisions happened here, not 100% sure what it's supposed to say--> was using the same generic type as `item`. We could do so inline:
 
 ```typescript
-async function logTheValue<ItemT>(item: ItemT): {loggedValue: string, original: ItemT, err: Error | undefined} {
+async function logTheValue<ItemT>(item: ItemT): Promise<{loggedValue: string, original: ItemT, err: Error | undefined}> {
```
crutchcorn: Ahh, my bad. Let's add a comment explaining why there's a promise here:

```suggestion
// Because this is an `async` function, we want to wrap the returned type value in a Promise
async function logTheValue<ItemT>(item: ItemT): Promise<{loggedValue: string, original: ItemT, err: Error | undefined}> {
```

content/blog/typescript-type-generics/index.md
```diff
@@ -199,14 +199,14 @@ interface LogTheValueReturnType<originalT> {
   err: Error | undefined;
 }
 
-async function logTheValue<ItemT>(item: ItemT): LogTheValueReturnType<ItemT> {
+async function logTheValue<ItemT>(item: ItemT): Promise<LogTheValueReturnType<ItemT>> {
```
crutchcorn: And I think we can even mention this now:

```suggestion
// Notice how we're even wrapping that interface in a built-in type with a generic argument for `Promise`!
async function logTheValue<ItemT>(item: ItemT): Promise<LogTheValueReturnType<ItemT>> {
```

content/blog/typescript-type-generics/index.md
```diff
@@ -311,7 +311,7 @@ interface TimestampReturn<T> {
 	isFuture: boolean;
 	obj: T
 }
-const checkTimeStamp(obj: T extends {time: Date}): TimestampReturn<T> {
+const checkTimeStamp = function(obj: T extends {time: Date}): TimestampReturn<T> {
```
crutchcorn:
```suggestion
const checkTimeStamp = (obj: T extends {time: Date}): TimestampReturn<T> => {
```

content/blog/typescript-type-generics/index.md
```diff
@@ -182,15 +182,15 @@ async function logTheValue(item) {
 }
 ```
 
-If we wanted to type the `logTheValue` function, we'd want to make sure to use a generic for the `loggedValue` was using the same generic type as `item`. We could do so inline:
+If we wanted to add typings for the `logTheValue` function, we'd want to make sure to use a generic for the `loggedValue` <!--Looks like some sentence revisions happened here, not 100% sure what it's supposed to say--> was using the same generic type as `item`. We could do so inline:
```
crutchcorn:
```suggestion
If we wanted to type the `logTheValue` function, we'd want to make sure to use a type generic for the input parameter `item`. By doing so, we could use that same geneeric for the return prop of `loggedValue` to ensure they both have the same typing. To do this, we could do so inline:
```

Reads a lot better

content/blog/typescript-type-generics/index.md
adueppen: Ok yeah I get that for the `console.log` thing. The emoji was just a bit confusing since at least on linux in FF it only appears in monochrome even though the other one gets color.

content/blog/typescript-type-generics/index.md
```diff
@@ -191,7 +191,7 @@ async function logTheValue<ItemT>(item: ItemT): Promise<{loggedValue: string, or
 }
 ```
 
-Alternatively, we could utilize another feature of generics: the ability to pass the type value of the generic manually, and make an interface with a generic and do so there:
+Alternatively, we could utilize another feature of generics: the ability to pass the type value of the generic manually by making an interface with a generic and doing so there:
```
crutchcorn:
```suggestion
Alternatively, we could utilize another feature of generics — the ability to pass the type value of the generic manually — and make an interface with a generic and do so there:
```

This better preserves the original sentence

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+
+However, unions have some limitations. You'll find that this doesn't give the example you might want:
+
+```typescript
+// ❌ This will yield an error
+// > Operator '+' cannot be applied to types '4' and 'string | number'.
+const newNumber = shouldBeNumber + 4;
+```
+
+The reason that the operation `shouldBeNumber + 4` yields this error is because it doesn't know which type to give you given your input, since the output is also explicitly a union (meaning that it could be a string OR a number returned — one of them supporting numeric operations, one of them not).
```
evelynhathaway: This is kinda hard to read. Maybe split this up differently and use "TypeScript doesn't know" and the variable name to reference existing subjects instead?

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+
+### Potential Solutions Disclaimer {#silly-examples-disclaimer}
+
+> Author's note:
+>
+> If you were using unions in your property definitions and left your return type blank, TypeScript would be able to infer what the return type should be just fine.
+>
+> That said, we're trying to build on concepts, so we're trying to provide some examples of where this might be used and what it does. There are also instances, such as type definition files, where this inference might not be available to an author of typings, as well as other limitations with this method that we'll see later,
```
evelynhathaway:
```suggestion
> That said, we're trying to build on concepts, so we're trying to provide some examples of where this might be used and what it does. There are also instances, such as type definition files, where this inference might not be available to an author of typings, as well as other limitations with this method that we'll see later.
```

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+
+```typescript
+returnProp({}) // Argument of type '{}' is not assignable to parameter of type 'string'.
+```
+
+This may seem obvious from the typings, but _we ideally want `returnProp` to accept ANY type because **we aren't using any operations that require knowing the type**._ (no addition or subtraction, requiring a number; no string concatenation that might restrict an object from being passed).
+
+## Potential Solution 3: Any {#generic-usecase-setup-any-solution}
+
+Of course, we could use the `any` in order to force any input and return type. (Goodness knows I've had my fair share of typing frustrations that ended with a few `any`s in my codebase!)
```
evelynhathaway:
```suggestion
Of course, we could use the `any` type to force any input and return type. (Goodness knows I've had my fair share of typing frustrations that ended with a few `any`s in my codebase!)
```

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+returnedObject.test();
+// This will exist, and TS will know it as a number
+returnedObject.objProperty;
+```
+
+> Author's note:
+>
+> The type variable does not need to be called `T`. In fact, while it seems to be commonplace for the community to use single-letter type variable names (often due to the length and complexity of the typings), there are many reasons why more explicit type names should be used.
+>
+> Remember, type variables are like other variables in that you need to maintain them and understand what they're doing in your code.
```
evelynhathaway: I love this. Thanks for adding this touch. 🙏🙏🙏

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+
+# Okay, but Why? {#logger-example}
+
+Why might we want to do this? [Returning an item as itself in a identity function](#generic-usecase-setup) is cool, but it's not very useful in its current state. That said, there ARE many many uses for generics in real-world codebases.
```
evelynhathaway:
```suggestion
Why might we want to do this? [Returning an item as itself in an identity function](#generic-usecase-setup) is cool, but it's not very useful in its current state. That said, there ARE many, many uses for generics in real-world codebases.
```

evelynhathaway: Are the all-caps words necessary throughout the text?

All-caps text is harder for people with reading disabilities, screen readers often read out each letter separately, and it's an interesting style choice that seems firmer than I would expect from a technical piece that can use rich text to convey the point just as strongly.

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+async function logTheValue(item) {
+	const jsonString = JSON.stringify(item, null, 2);
+
+	let err = undefined;
+
+	try {
+		// Attempt to write a new log file. If this fails, save the error to the `err` variable
+		await writeFileAsync(`/logs/${Date.now()}`, jsonString);
+	} catch (e) {
+		err = e;
```
evelynhathaway: The manual "hoisting" of `e` to `err` may not be the easiest for some, especially with two shortenings rather than verbose names. Not sure what's best to do here, but in general I like being explicit since `e` is often `event` and `err` saves only two letters.

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+  err: Error | undefined;
+}
+
+// Notice how we're even wrapping that interface in a built-in type with a generic argument for `Promise`!
+async function logTheValue<ItemT>(item: ItemT): Promise<LogTheValueReturnType<ItemT>> {
+	// ... Function body here
+}
+```
+
+WIth these few features, we're able to utilize much of the functionality of generics.
```
evelynhathaway:
```suggestion
With these few features, we're able to utilize much of the functionality of generics. 
```

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+}
+
+interface ImageConvertMethods<DataType> {
+  // This is the typing of a method. It will take a prop of the generic type and return the generic type
+  toPNG: (data: DataType) => DataType;
+  toJPG: (data: DataType) => DataType;
+}
+
+
+type ImageTypeWithConvertMethods<DataType> = ImageType<DataType> & ImageConvertMethods<DataType>
```
evelynhathaway: I've noticed some inconsistent indenting in some of these code blocks. I don't have my editor up at the moment, but maybe some find and replace regex would fix this baby right up!

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+```
+
+# Okay, but why-_er_? {#polymorphic-functions}
+
+Well, "I want to see your credentials" person you, type generics enable us to do things like provide typings for **polymorphic functions**. _Polymorphic functions are functions that can accept a myriad of different types and handle them differently._
```
evelynhathaway: I understand you're jokingly referring to the reader, but the particular phrasing `"I want to see your credentials" person you` is confusing because there are three subjects that I can't quite piece together.

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+
+> Polymorphic functions are not unique to TypeScript, the things learned here about polymorphic functions can be applied to other languages as well. They also provide some real-world insight to usages of generics and when they might be able to be used.
```
evelynhathaway:
```suggestion
> Polymorphic functions are not unique to TypeScript; the things learned here about polymorphic functions can be applied to other languages as well. They also provide some real-world insight into the usages of generics and when they could be used.
```

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+
+For example, let's take a look at the code for the `toPNG` :
```
evelynhathaway:
```suggestion
For example, let's take a look at the code for the `toPNG`:
```

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+		const imgBuffer = getBufferFromBaseStr(data);
+		const pngBuffer = convertBufferToPNG(imgBuffer);
+		return bufferToBase64(pngBuffer);	
+	} else {
+		throw 'toPNG only accepts arrays, buffers, or strings'
+	}
+}
+```
+
+Even though this function accepts various data types, it handles them differently under the hood! Functions that have this type of accept-many-handle-each-slightly-differently behavior are called **Polymorphic Functions**. They're particularly useful in utility libraries.
```
evelynhathaway: The comma is too important to remove in this longer phrase in my opinion.

```suggestion
Even though this function accepts various data types, it handles them differently under the hood! Functions that have this type of "accept many, handle each slightly differently" behavior are called **Polymorphic Functions**. They're particularly useful in utility libraries.
```

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+
+```
+
+In this case, we can rely on implicit type casting to ensure that we're able to pass `{time: new Date()}` but not `{}` as values for `obj`.
+
+# Conclusion
+
+And that's all I have for generics! Their usages are far and wide, and now you're able to apply your knowledge in code! We're hoping to have more posts on TypeScript soon - both more introductory and advanced. 
+
+Questions? Feedback? Sound off in the comments below, we'd love to hear from you!
```
evelynhathaway:
```suggestion
Questions? Feedback? Sound off in the comments below; we'd love to hear from you!
```

content/blog/typescript-type-generics/index.md
crutchcorn: I wanted to make sure to bring it up since I think it's important. I want to make sure in more complex articles to move away from  `T` as well, but I think in an intro to generics seeing `T` might help understand conventions and keep an idea of abstract type concepts

content/blog/typescript-type-generics/index.md
crutchcorn: You're absolutely right. I should be using bolded text here:

```suggestion
Why might we want to do this? [Returning an item as itself in an identity function](#generic-usecase-setup) is cool, but it's not very useful in its current state. That said, there **are** many, many uses for generics in real-world codebases.
```

content/blog/typescript-type-generics/index.md
crutchcorn: `error` can be preserved in some instances so I wanted to remove it. I'll add a comment to make it more clear but I don't think I want to change the code too much here

content/blog/typescript-type-generics/index.md
```diff
@@ -0,0 +1,340 @@
...
+ */
+async function logTheValue(item) {
+	const jsonString = JSON.stringify(item, null, 2);
+
+	let err = undefined;
+
+	try {
+		// Attempt to write a new log file. If this fails, save the error to the `err` variable
+		await writeFileAsync(`/logs/${Date.now()}`, jsonString);
+	} catch (e) {
```
crutchcorn:
```suggestion
      // Catch any errors and keep them as the `e` variable to assign to `err` later
	} catch (e) {
```

content/blog/typescript-type-generics/index.md
crutchcorn: Yeah, I'll be fixing the formatting of the code blocks in the next commit to use tabs as usual :)

content/blog/typescript-type-generics/index.md
crutchcorn: Would `Well, you "I want to see your credentials" person you` help?

I wanted to avoid doing `Mr.` or `Ms.`/etc as that opens a frustrating can of assumed gender worms. I might just remove this problem all together and rephrase it

content/blog/typescript-type-generics/index.md
crutchcorn:
```suggestion
The reason that the operation `shouldBeNumber + 4` yields this error is because you've told TypeScript that `shouldBeNumber` is either a number **or** a string by making the output explicitly typed as a union. As a result, TypeScript is unable to do addition between a number and a string (which is one of the potential values) and therefore throws an error.
```

content/blog/typescript-type-generics/index.md
crutchcorn: How's this?

```suggestion
My my, you don't seem to take my word for it when I tell you that type generics are useful. That's alright, I suppose; After all, doubt while learning can lead to some great questions! 😉

Type generics enable us to do things like provide typings for **polymorphic functions**. _Polymorphic functions are functions that can accept a myriad of different types and handle them differently._
```

crutchcorn:
@evelynhathaway your edits (as always) are amazing and MASSIVELY appreciated.

That said, 4AM nights aren't healthy for anyone - your health comes first take care of yourself 💜

(for those not familiar w/ Evelyn and my mental mapping of heart colors - the purple heart is meant for friendship and support)

PR #98, opened 2019-10-09
Typescript intro post

crutchcorn:


content/blog/introduction-to-typescript/index.md
```diff
@@ -0,0 +1,212 @@
...
+
+```javascript
+function addFive(input) {
+    return input + 5;
+}
+```
+
+If you keep `addFive`'s `input` parameter without an explicit type, it will try to do its best to detect the type based on the operations you run on the value. **While these inferred types are often very useful, it often has difficulty doing so in an accurate manor**. _This is why manually assigning types is almost always preferred to leaving them be inferred_. If it couldn't properly detect a type for the value in this example, it does nothing to prevent strings from being passed as the parameter value.
+
+Although examples like this are simple, strict typings can also become fairly complex ([see the Advanced section of the official handbook for such examples](https://www.typescriptlang.org/docs/handbook/advanced-types.html)) in order to maintain maximum type strictness.
```
evelynhathaway:
```suggestion
Although examples like this are simple, strict typings can also become fairly complex to maintain maximum type strictness. See the [Advanced section of the official handbook](https://www.typescriptlang.org/docs/handbook/advanced-types.html) for examples that illustrate this.
```

content/blog/introduction-to-typescript/index.md
```diff
@@ -0,0 +1,212 @@
...
+Also, because typings do nothing to test against the logic of your program, they should not be seen as a replacement for testing, but rather a companion to them. With strict typings and proper testing, regressions can be severely limited and improve code quality of life.
```
evelynhathaway: Should this be promoted with styling or position within the heading? This was very important for me when first starting out with TS.

content/blog/introduction-to-typescript/index.md
```diff
@@ -0,0 +1,212 @@
...
+
+##### Typing Mishaps Happen {#typings-can-be-wrong-too}
+
+Remember, because typings are kept separately from the project's logic code, typings can be misleading, incomplete, or otherwise incorrect. While this can also happen with TypeScript logic code, it tends to be more actively mitigated as a project's ability to compile (and therefore redistribute) relies on that typing information. This isn't to say that you should immediately mistrust typings, but simply a reminder that they too may have their flaws - just as any other codebase.
```
evelynhathaway:
```suggestion
Remember, because typings are kept separately from the project's logic code, typings can be misleading, incomplete, or otherwise incorrect. While this can also happen with TypeScript logic code, it tends to be more actively mitigated as a project's ability to compile (and therefore distribute) relies on that typing information. This isn't to say that you should immediately mistrust typings, but this is simply a reminder that they too may have their flaws — just as any other part of a codebase.
```
- I'd use `distribute` or `deploy` here as they are the primary distributors.
- Minor wording

content/blog/introduction-to-typescript/index.md
```diff
@@ -0,0 +1,212 @@
...
+
+## Don't Forget To Document {#typescript-is-not-documentation}
+
+Just as typings shouldn't replace tests, typings should also not replace documentation or comments. Typings can help understand what inputs and outputs you're expecting, but just the same as with testing; _it doesn't explain what the logic does or provide context as to why the data types have specific properties_, what the properties are used for, and so forth. Additionally, typings often do little to help explain how to contribute in the larger scale or the design principles that may be applied in the design of the application.
```
evelynhathaway: > Additionally, typings often do little to help explain how to contribute in the larger scale or the design principles that may be applied in the design of the application.

I'm unclear on what this is talking about. I think minor rephrasing would help make this point land.

content/blog/introduction-to-typescript/index.md
```diff
@@ -0,0 +1,212 @@
...
+
+Essentially, I just want to make sure to iterate that while there may be tools that can let you auto-document the properties and the property type from TypeScript annotations, hand-done documentation is still extremely important
```
evelynhathaway:
```suggestion
Essentially, I just want to make sure to iterate that while there may be tools that can let you auto-document the properties and the property type from TypeScript annotations, hand-done documentation is still extremely important.
```

content/blog/introduction-to-typescript/index.md
```diff
@@ -0,0 +1,212 @@
...
+```
+
+```typescript
+// index.d.ts
+declare function ANumberToAString(numProp: number): string; // Accept a number arg, return a string
+```
+
+##### Community Hosting {#definitely-typed}
+
+Additionally, because TypeScript has a well established and widely used install-base, **there are already many different definition files in the wild for supporting non-TypeScript supporting projects**. One of the more extensive collections of these typings lives at the [DefinitelyTyped repository](https://github.com/DefinitelyTyped/DefinitelyTyped), which publishes the typings of their typings under the package names `@types/your-package-name` (where `your-package-name` is the name of the project you're looking for typings of) that you can look for on your package manager.
```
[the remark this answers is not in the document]
crutchcorn:
```suggestion
Additionally, because TypeScript has a well established and widely used install-base, **there are already many different definition files in the wild for supporting non-TypeScript supporting projects**. One of the more extensive collections of these typings lives at the [DefinitelyTyped repository](https://github.com/DefinitelyTyped/DefinitelyTyped), which publishes the package's community typings under the package names `@types/your-package-name` (where `your-package-name` is the name of the project you're looking for typings of) that you can look for on your package manager.
```

content/blog/introduction-to-typescript/index.md
crutchcorn:
```suggestion
**Also, because typings do nothing to test against the logic of your program, they should not be seen as a replacement for testing, but rather a companion to them.** _With strict typings and proper testing, regressions can be severely limited and improve code quality of life._
```

content/blog/introduction-to-typescript/index.md
crutchcorn:
```suggestion
Just as typings shouldn't replace tests, typings should also not replace documentation or comments. Typings can help understand what inputs and outputs you're expecting, but just the same as with testing; _it doesn't explain what the logic does or provide context as to why the data types have specific properties_, what the properties are used for, and so forth. Additionally, typings often do little to help explain how to contribute in the larger scale when talking about documentation. For example, in a large-scale application, there may be come complex data patterns or order-of-operations that are required to do a task. Typings alone will not effectively communicate these design principles that are integral to the usage of the code.
```

content/blog/introduction-to-typescript/index.md
```diff
@@ -0,0 +1,212 @@
...
+```typescript
+let numberHere: number = 0;
+// 🛑 Error will be thrown during compilation:
+// Type '"Test"' is not assignable to type 'number'.
+numberHere = 'Test';
+
+const newNumber = 10 - numberHere;
+```
+
+This point is made even more complex when dealing with how both are constructed and handed mentally or dealing with classes. While [ES6 classes are syntactical sugar on top of JavaScript's prototype system](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) and there [are only seven base-types](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures) (called _primitive types_) in JavaScript that every other value is composed of, TypeScript has a much more robust type system.
```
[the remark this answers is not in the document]
crutchcorn:
```suggestion
This point is made even more complex when dealing with how both values are handled internally in JavaScript. Especially when talking about ES6 classes, _the type-strict nature of TypeScript types and the "types" that are understood in JavaScript are not the same_. While [ES6 classes are syntactical sugar on top of JavaScript's prototype system](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) and there [are only seven base-types](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures) (called _primitive types_) in JavaScript that every other value is composed of, TypeScript has a much more robust type system.
```

content/blog/introduction-to-typescript/index.md
```diff
@@ -205,8 +210,6 @@ Essentially, I just want to make sure to iterate that while there may be tools t
 
 And with that, we have a better understanding of what TypeScript is! I hope this has been informative and helpful for those that may be new to the language in particular. What'd you learn, let us know!
 
-Now that you're more familiar with TypeScript, maybe you'd like to play around with one of their more experienced functionality: [Type generics](https://unicorn-utterances.com/posts/typescript-type-generics/)?
-
-We have a whole post around that concept as well, [you can find that here](https://unicorn-utterances.com/posts/typescript-type-generics/).
+Now that you're more familiar with TypeScript, maybe you'd like to play around with one of their more experienced functionality: [Type generics](https://unicorn-utterances.com/posts/typescript-type-generics/)? We have a whole post around that concept as well, [you can find that here](https://unicorn-utterances.com/posts/typescript-type-generics/).
```
evelynhathaway: Great! These links are the same, is that duplication intentional?

crutchcorn: Yes'm it is :) I wanted to include just in case

Issue #70, opened 2019-09-21
Switch to Next.JS SSR rather than compile-time render


crutchcorn:
We're finding more and more usecases where we'd want to do things with non-compile-time SSR. It aligns much better with our long-term goals and use-cases and as a result would be worth the engineering offset

For our initial developmental effort, we could have a small Node express app that enables us to still load data from our Markdown and JSON files to present a very similar dev experience to what we have now:

https://graphql.org/code/#express-graphql-graphql-js-running-an-express-graphql-server-github-https-github-com-graphql-express-graphql-npm-https-www-npmjs-com-package-express-graphql

Luckily, because of the nature of Gatsby plugins, it would be relatively trivial to lookup how they're doing it and either:

1) Provide a compat layer between our custom NextJS instance

> I like this idea a lot personally, it wouldn't be too bad to run the lifecycles in the same expected lifecycle methods

> There may be some problems matching up lifecycle methods due to the lack of compile-time SSR, but that'll be something we'll have to engineer around and discuss at the time (plus, we can probably have an overwrite to what lifecycle method things run at) 

2) Port their plugins to be entirely in-house

> While this would be much faster, likely, it would be bad due to note contributing back to the broader community, duplicating dev effort, etc

This will need to be a larger discussion from a technical perspective, feedback encouraged. Once a path forward is more standardized, we can break this ticket out to further other tickets to break out the developmental efforts

Potentially impacted by this: #65, #60

crutchcorn:
Something worth tracking is NextJS official CSS module support: 

https://nextjs.org/blog/next-9-1#built-in-css-support

crutchcorn:
Closing this ticket as "Out-of-scope" for the present. We may revisit this in the future

crutchcorn:
This was actually completed with:

https://github.com/unicorn-utterances/unicorn-utterances/pull/289

PR #107, opened 2019-10-22
Pagination support

crutchcorn:
I have a very early version of pagination working, but am limited in pushing it or finalizing it by two features:

- Search
- Filter

I have also yet to add pagination support to author pages yet

Because both of the features currently rely on filtering the list of what's present on the page memory, however the way I'm doing things right now by limiting the graphQL call by using `skip` and `limit`. What we'll need to do is multi-fold:

- [x] Move initial posts shown to JS logic rather than graphql
    This will allow us to do things like changing the items that're shown when the user sorts AFTER using the page results
    - [x] Make sure this is utility based and able to migrate to author pages as well using the same utility
- [x] Implement pagination support during an active filter or search (this will be client-side only. No need to worry about `createPage`)
    - [x] Logic should probably live in a `PostListHandler` component which listens to a Context that has the filter and search results and handles pagination from there
    - [x] When no search or filter are applied, switch back from using client-side sort handling to using the `navigate` method
- [x] In order to avoid memory bloat, the query for the `post-list` page will need to be limited in scope. 
     What I mean by this is that we can't simply use `PostInfo` for the query any further. Another fragment will likely need to be provided in order to have a `PostInfoListDisplay` which we can keep in memory in order to do client-side sort and filtering

One thing that I don't/can't worry about it handling URL changes during search + pagination. This overhead would REQUIRE the user have JavaScript and be extremely challenging to manage. As a result, this behavior will be a known potential improvement with a `DONTFIX` result until the migration to NextJS SSR per #70


crutchcorn:
Note for @unicorn-utterances/devops - we also need to make a 301 from `/pages/1` to `/`

crutchcorn:
It should also be called out that this seems to compound to the "Go back" button issue mentioned #45.

crutchcorn:
This PR is now done and ready to test. A few notes about testing:

- You'll probably want to change the `postsPerPage` value in the `gatsby-node.js` file to test actual pagination

- Please test filter and search as well. Remember that the page URL may not match. This is known and will not be fixed until after #70 is finished

- Please test both the main page and the author page

adueppen:
So far taking a look at the functionality it seems good, although something that we might want to work on is keeping the page buttons in the same location (whether the next/previous buttons are there or not) since that makes it easier to navigate between pages for me at least.

crutchcorn:
I think make this a fast follow and do so in the next PR

evelynhathaway:
@crutchcorn is the redirect set up?

crutchcorn:
Not yet. I can do so tonight

PR #112, opened 2019-10-26
Fix syntax highlighting

adueppen:
I noticed a few small issues that seemed to be happening with syntax highlighting and code blocks in general, including a warning about prismjs not supporting Gradle yet. Aliasing it to Groovy manages to at least provide some basic highlighting.


adueppen:
I tried messing around with the `gatsby-remark-prismjs` a bit and unfortunately it seems that the warning mentioned in the second commit is unavoidable, even setting `noInlineHighlight` to `true` (which also messes with the styling of inline code) still produced the warning.

PR #113, opened 2019-10-26
Make sure the proper profile pic size is used in UserProfilePic

adueppen:
I'm not sure if using a boolean like this is the best approach but it works well enough. This also changes the smallPic size to 68px instead of 60px.

### DIFF
--- a/src/components/layout/layout.js
+++ b/src/components/layout/layout.js
@@ -86,7 +86,7 @@ export const authorFragmentQuery = graphql`
     }
     profileImg {
       childImageSharp {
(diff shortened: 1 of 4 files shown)


adueppen:
Note that `post-list.js` will also have to be changed in PR #107 in order for the small pic to be the correct size.

crutchcorn:
Can you rebase off of intergration to get the changes in #107 as you mentioned?

crutchcorn:
Closing, as design and changes have been modified

PR #116, opened 2019-11-04
Added prettier to SCSS files

crutchcorn:
We currently have an issue where upon trying to commit, we're unable to do so due to due to SCSS errors being thrown. In order to fix this long-term, we added stylelint to style SCSS for prettier as well (As originally suggested by @adueppen )


crutchcorn:
Works for me 👍

PR #117, opened 2019-11-06
Fix disqus slug

crutchcorn:
I'll have to migrate the Disqus posts, but we were resetting the disqus identifier by the theme that was used to load it

https://sporadicallyyours.com/techbytes/2013/5/19/change-url-without-losing-comments


evelynhathaway:
I'm curious to why the workaround was made to use the theme for Disqus

crutchcorn:
@evelynhathaway the problem comes down to Disqus basing the theme color off of the background color when it initializes. I mistakenly read the docs as if this was the `key` value (which is what `react` uses to re-render an element) and changing both the unique ID and the `key` at the same time. The `key` is the real fix for the theme toggle but I was changing too much at the same time and wrongly attributed the fix to the Disqus ID

PR #121, opened 2019-11-20
"Understanding the DOM" post

crutchcorn:


content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+
+There's also the idea of a _"sibling"_ node. When a node's parent has more than one child, those other nodes are that child node's _"siblings"_.
+
+![A chart showing the relationships between parents and siblings](./dom_relationships.svg)
+
+Altogether, the terminologies used to refer between these node "leaves" are extremely similar to the terminology often used with family trees.
+
+There are some rules to this tree that's created from these "nodes":
+
+- There must be one "root" or "trunk" node, there cannot be more than one root
```
evelynhathaway:
```suggestion
- There must be one "root" or "trunk" node, and there cannot be more than one root
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+<main id="a">
+	<p id="e"></p>
+  <ul id="b">
+    <li id="c"></li>
+    <li id="d"></li>
+  </ul>
+</main>
+```
+
+The browser is able to keep in mind, while it's moving through the tree, that it needs to find an element with the `ID` of `b` and then mark it's `li` children with a red background. They're "children" because the DOM tree keeps that relationship info that's defined by the HTML.
```
evelynhathaway:
```suggestion
The browser is able to keep in mind, while it's moving through the tree, that it needs to find an element with the `ID` of `b` and then mark it's `<li>` children with a red background. They're "children" because the DOM tree keeps that relationship info that's defined by the HTML.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+
+![A chart showing the 'ul' tag highlighted in green with the children 'li' tags marked in red](./dom_tree_with_css.svg "Diagram showing the above code as a graph")
+
+> The `ul` element is marked as green just to showcase that it is the element being marked by the first part of the selector
```
evelynhathaway:
```suggestion
> The `<ul>` element is marked as green just to showcase that it is the element being marked by the first part of the selector
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+
+Typically, the browser will "visit" it's nodes in a specific order. For example, in the above chart, the browser might start at the `main` tag, then go to the `p` tag, then visit the `ul` tag, and finally the two children in order from left-to-right (`li id="c"` , `li id="d"`).
```
evelynhathaway:
```suggestion
Typically, the browser will "visit" it's nodes in a specific order. For example, in the above chart, the browser might start at the `<main>` tag, then go to the `<p>` tag, then visit the `<ul>` tag, and finally the two children in order from left-to-right (`<li id="c">` , `<li id="d">`).
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+
+The browser, knowing what CSS to look for, is able to see the `ul` with the correct ID and know to mark it's children with the correct metadata that matches the selector with the relevant CSS.
```
evelynhathaway:
```suggestion
The browser, knowing what CSS to look for, is able to see the `<ul>` with the correct ID and know to mark its children with the correct metadata that matches the selector with the relevant CSS.
```

content/blog/understanding-the-dom/index.md
evelynhathaway: Even though the saying "`<tag>` tag" may sound like saying ATM machine, I believe it's still more clear, even this far into the document.

In case I miss some in my GitHub-side review, it may be nice to go back and double check them with
```
`\w+`
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+
+This tree relationship also enables CSS selectors such as [general sibling selector (`~` )](https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator) or the [adjacent sibling selector (`+`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator) to find siblings to a given selector
```
evelynhathaway:
```suggestion
This tree relationship also enables CSS selectors such as the [general sibling selector (`~`)](https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator) or the [adjacent sibling selector (`+`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator) to find siblings to a given selector.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+
+![A showcase of the above selectors and how they always look forward, never behind](css_selectors_demo.svg)
+
+> Interestingly, one of the questions that I've often heard asked is a "parent selector". The idea behind this is that the [direct child selector (`>`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator)  exists, why not have the ability to mark any parent of `.classname` selectors.
```
evelynhathaway:
```suggestion
> Interestingly, one of the questions that I've often heard asked concerns a "parent selector". The idea behind the question is that the [direct child selector (`>`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator) exists, so why not have the ability to mark any parent of `.classname` selectors?
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+
+This tree relationship also enables CSS selectors such as [general sibling selector (`~` )](https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator) or the [adjacent sibling selector (`+`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator) to find siblings to a given selector
+
+![A showcase of the above selectors and how they always look forward, never behind](css_selectors_demo.svg)
+
+> Interestingly, one of the questions that I've often heard asked is a "parent selector". The idea behind this is that the [direct child selector (`>`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator)  exists, why not have the ability to mark any parent of `.classname` selectors. 
+>
+> The answer behind that is: Performance. The [W3 organization](https://www.w3.org/Style/CSS/#specs) (the organization who maintains the HTML and CSS standard specification) points to the tree structure of the DOM and the algorithm behind how browsers traverse the DOM (or, "visit" the nodes in order to figure out what CSS to apply)  as not being performant when allowing parent selectors.
```
evelynhathaway:
```suggestion
> The answer behind that is performance. The [W3 organization](https://www.w3.org/Style/CSS/#specs) (the organization who maintains the HTML and CSS standard specification) points to the tree structure of the DOM and the algorithm behind how browsers traverse the DOM (or, "visit" the nodes in order to figure out what CSS to apply)  as not being performant when allowing parent selectors.
```
- `is` is functionally equivalent to the colon here
  - You can also replace `is` with the colon (and lower casing `Performance`) or use an ellipsis if the pause is important

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+>
+> The reason behind this is that browsers read from top-down in the DOM and apply CSS as they find matching nodes; CSS doesn't command the browser to do anything to the DOM, but rather provides the metadata for the DOM to apply the relevant CSS when the browser comes across that specific node.
+>
+> As mentioned before, they start at the root node, keep notes on what they've seen, then move to children. Then, they move to siblings, etc. Specific browsers may have slight deviations on this algorithm, but for the most part they don't allow for upwards vertical movement of nodes within the DOM.
```
evelynhathaway:
```suggestion
> As mentioned before, they start at the root node, keep notes on what they've seen, then move to children. Then, they move to siblings, etc. Specific browsers may have slight deviations on this algorithm, but for the most part, they don't allow for upwards vertical movement of nodes within the DOM. 
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+>
+> The reason behind this is that browsers read from top-down in the DOM and apply CSS as they find matching nodes; CSS doesn't command the browser to do anything to the DOM, but rather provides the metadata for the DOM to apply the relevant CSS when the browser comes across that specific node.
+>
+> As mentioned before, they start at the root node, keep notes on what they've seen, then move to children. Then, they move to siblings, etc. Specific browsers may have slight deviations on this algorithm, but for the most part they don't allow for upwards vertical movement of nodes within the DOM. 
+
+
+
+# Using The Correct Tags {#accessibility}
+
+HTML, as a specification, has tons of tags that are able to be used at one's disposal. These tags contain various metadata internally to the browser to provide information about what you want to render in the DOM. This metadata can then be handled by the browser how it sees fit; it may apply default CSS styling, it may change the default interaction the user has with it, or even what behavior that element has upon clicking on it (in the case of a button in a form).
```
evelynhathaway:
```suggestion
HTML, as a specification, has tons of tags that can be used at one's disposal. These tags contain various metadata internally to the browser to provide information about what you want to render in the DOM. This metadata can then be handled by the browser how it sees fit; it may apply default CSS styling, it may change the default interaction the user has with it, or even what behavior that element has upon clicking on it (in the case of a button in a form). 
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+
+Some of these tag defaults are part of the specification while others are left up to the browser vendor to decide. This is why, in many instances, developers may choose to use something like `normalize.css` in order to set all of the element CSS defaults to an explicit set of defaults; Doing so can avoid having the UI of a webpage look different from browser-to-browser thanks to deviations on default CSS styling on specific tags.
+
+This metadata is also why it's so important that your application utilizes the expected HTML tags and not simply default to `div`s with CSS applied to simulate other items. Two of the biggest advantages to responsibly utilizing the metadata system the browser has built into it by using the correct tags are SEO and accessability.
```
evelynhathaway:
```suggestion
This metadata is also why it's so important that your application utilizes the expected HTML tags and not simply default to `<div>`s with CSS or JavaScript applied to simulate other items. Two of the biggest advantages to responsibly utilizing the metadata system the browser has built into it by using the correct tags are search engine optimization (SEO) and accessibility.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+
+
+
+# Using The Correct Tags {#accessibility}
+
+HTML, as a specification, has tons of tags that are able to be used at one's disposal. These tags contain various metadata internally to the browser to provide information about what you want to render in the DOM. This metadata can then be handled by the browser how it sees fit; it may apply default CSS styling, it may change the default interaction the user has with it, or even what behavior that element has upon clicking on it (in the case of a button in a form). 
+
+Some of these tag defaults are part of the specification while others are left up to the browser vendor to decide. This is why, in many instances, developers may choose to use something like `normalize.css` in order to set all of the element CSS defaults to an explicit set of defaults; Doing so can avoid having the UI of a webpage look different from browser-to-browser thanks to deviations on default CSS styling on specific tags.
```
evelynhathaway:
```suggestion
Some of these tag defaults are part of the specification, while others are left up to the browser vendor to decide. This is why, in many instances, developers may choose to use something like [`normalize.css`](https://github.com/necolas/normalize.css/) to set all of the element CSS defaults to an explicit set of defaults. Doing so can avoid having the UI of a webpage look different from browser-to-browser thanks to deviations on default CSS styling on specific tags.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+
+```html
+<div>
+	<div>Bananas</div>
+	<div>Apples</div>
+	<div>Oranges</div>
+</div>
+```
+
+In this example, your browser only knows that you're looking to display text on screen. If a user utilizing a screen reader reaches the site, the browser doesn't know that it should inform the user that there are three items in a list (something that sight-impaired users would greatly value to know, in order to tab through the list effectively) as you've done nothing to inform the user that it is a list of items: Only that it's a set of `div` generic containers.
```
evelynhathaway:
```suggestion
In this example, your browser only knows that you're looking to display text on-screen. If someone utilizing a screen reader reaches the site, the browser doesn't know that it should inform them that there are three items in a list (something that people with low vision would greatly value to know, in order to tab through the list effectively) as you've done nothing to inform the user that it is a list of items: only that it's a set of `<div>` generic containers. 
```

- When talking about a11y, I especially like focusing on people, not users
- I avoid the word `impairment` when speaking about a disablility
  - This isn't the most common language change, but one I believe makes a difference

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+<div>
+	<div>Bananas</div>
+	<div>Apples</div>
+	<div>Oranges</div>
+</div>
+```
+
+In this example, your browser only knows that you're looking to display text on screen. If a user utilizing a screen reader reaches the site, the browser doesn't know that it should inform the user that there are three items in a list (something that sight-impaired users would greatly value to know, in order to tab through the list effectively) as you've done nothing to inform the user that it is a list of items: Only that it's a set of `div` generic containers. 
+
+Likewise, when Google's robots walk through your site, they won't be able to parse that you're displaying lists to your users. As a result, your search rating for "list of best places" might be impacted, since the site doesn't appear to contain any list at all
```
evelynhathaway:
```suggestion
Likewise, when Google's robots walk through your site, they won't be able to parse that you're displaying lists to your users. As a result, your search rating for "list of best places" might be impacted since the site doesn't appear to contain any list at all.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
+---
+{
+	title: 'Understanding The Dom: How Browsers Show Things on Screen',
```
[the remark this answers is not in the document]
crutchcorn:
```suggestion
	title: 'Understanding The DOM: How Browsers Show Content on Screen',
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+	description: 'Learn how the browser internally handles HTML and CSS to show the user things on-screen',
+	published: '2019-11-26T22:12:03.284Z',
+	authors: ['crutchcorn'],
+	tags: ['dom', 'browser internals'],
+	attached: [],
+	license: 'cc-by-nc-sa-4'
+}
+---
+
+Any web application relies on some fundamental technologies: HTML, CSS, and JavaScript. Even advanced front-end JavaScript frameworks such as Angular, React, or Vue will utilize some level of HTML to load the JavaScript. This understood, how the browser handes HTML and CSS under-the-hood can be quite the mystery. With this article, I'm going to explain what the browser does under-the-hood to understand what to show to the user.
```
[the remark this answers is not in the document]
crutchcorn: Alternative way to write it would be "Knowing this" or alternative. It's not a perfect fit here I'll just change it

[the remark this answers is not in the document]
crutchcorn:
```suggestion
Any web application relies on some fundamental technologies: HTML, CSS, and JavaScript. Even advanced front-end JavaScript frameworks such as Angular, React, or Vue will utilize some level of HTML to load the JavaScript. That said, how the browser handles HTML and CSS under-the-hood can be quite the mystery. In this article, I'm going to explain what the browser does to understand what it should show to the user.
```

content/blog/understanding-the-dom/index.md
crutchcorn:
```suggestion
> The answer behind that: performance. The [W3 organization](https://www.w3.org/Style/CSS/#specs) (the organization who maintains the HTML and CSS standard specification) points to the tree structure of the DOM and the algorithm behind how browsers traverse the DOM (or, "visit" the nodes in order to figure out what CSS to apply)  as not being performant when allowing parent selectors.
```

content/blog/understanding-the-dom/index.md
crutchcorn: People vs users is an industry wide change I wouldn't mind being made, and it's a great shift from `impairment`. Thanks for raising this

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,521 @@
...
+
+# The DOM {#the-dom}
+
+Just as the source code of JavaScript programs are broken down to abractions that are more easily understood by the computer, so too is HTML. HTML, initially being derived from SGML (the basis for XML as well), actually _forms a tree structure in memory_ in order to know [how to lay things out and do various other tasks](#how-the-browser-uses-the-dom). This tree structure in memory is _called the Document Object Model_ (or _DOM_ for short).
```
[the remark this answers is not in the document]
crutchcorn: I'm not against changing this, but it reads oddly to me for some reason. Let's try:

```suggestion
Just as the source code of JavaScript programs are broken down to abstractions that are more easily understood by the computer, so too is HTML. HTML, initially being derived from [SGML (the basis for XML as well)](https://en.wikipedia.org/wiki/Standard_Generalized_Markup_Language), actually _forms a tree structure in memory_ in order to [describe the relationships, layout, and executable tasks for items in the tree](#how-the-browser-uses-the-dom). This tree structure in memory is _called the Document Object Model_ (or _DOM_ for short).
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,520 @@
...
+<ol>
+	<li>Bananas</li>
+	<li>Apples</li>
+	<li>Oranges</li>
+</ol>
+```
+
+In this example, both the browsers as well as Google's scraper bots are able to discern that this is a list with three list items within it.
+
+> While there ARE tags that may potentially  impact SEO somewhat significantly, it's unlikely `ul` and `li` would significantly impact your SEO scores.
```
evelynhathaway:
```suggestion
> While there **are** tags that may potentially  impact SEO somewhat significantly, it's unlikely `<ul>` and `<li>` would significantly impact your SEO scores.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,520 @@
...
+>
+> Needless to say, it's still good to use symantic (correctly tagged) HTML as there are users that use screen-readers that will benefit greatly from these minor changes. Additionally, it can make code more readable and parsable with automated tools
```
evelynhathaway:
```suggestion
> Needless to say, it's still good to use semantic (correctly tagged) HTML as people that use screen-readers and other assistive technologies benefit greatly from these minor changes. Additionally, it can make code more readable and parsable with automated tools.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,520 @@
...
+
+We're able to even add further metadata to an element by using attributes. For example, let's say that I want to add a title to the list to be read upon a screen reader gaining focus on the element, we could use the `aria-label` attribute:
```
evelynhathaway:
```suggestion
We're able to even add further metadata to an element by using attributes. For example, let's say that I want to add a title to the list to be read upon a screen reader gaining focus on the element; we could use the `aria-label` attribute:
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+```html
+<ol aria-label="My favorite fruits">
+	<li>Bananas</li>
+	<li>Apples</li>
+	<li>Oranges</li>
+</ol>
+```
+
+In fact, the default metadata that is defaulted by specific tags can be directly applied manually to an element of a different tag. The metadata that is passed to the browser when using `<li>` is typically involving that element pertaining to a listitem, using the `role` attribute, we can add that information to a `<div>` itself.
```
evelynhathaway:
```suggestion
In fact, the default metadata that is defaulted by specific tags can be directly applied manually to an element of a different tag. The metadata that is passed to the browser when using `<li>` is typically involving that element pertaining to a `listitem`, using the `role` attribute, we can add that information to a `<div>` itself.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+```html
+<ol>
+	<div role="listitem">Bananas</div>
+	<div role="listitem">Apples</div>
+	<div role="listitem">Oranges</div>
+</ol>
+```
+
+> It's worth mentioning that this example is generally considered malpractice. While you may have been able to preserve _some_ of the metadata from an `<li>` tag in a `<div>` element, it's extremely difficult to catch all of the defaults a browser might apply to the original tag that may enhanse a sight-impaired user's experience using screen-readers.
```
evelynhathaway:
```suggestion
> It's worth mentioning that this example is generally considered malpractice. While you may have been able to preserve _some_ of the metadata from a `<li>` tag in a `<div>` element, it's extremely difficult to catch all of the defaults a browser might apply to the original tag that may enhance the experience of someone that uses a screen-reader.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+>
+> This is all to say: unless you have a **really** good reason for using `role` rather than an approprate tag, stick with the related tag. Just as any other form of engineering, properly employing HTML requires nuance and logic to be deployed at the hand of the implementing developer.
```
evelynhathaway:
```suggestion
> This is all to say: unless you have a **really** good reason for using `role` rather than an appropriate tag, stick with the related tag. Just as any other form of engineering, properly employing HTML requires nuance and logic to be deployed at the hand of the implementing developer.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+	</p>
+</body>
+</html>
+```
+
+The `document` object has the ability to get the `<body>` node ([`document.body`](https://developer.mozilla.org/en-US/docs/Web/API/Document/body)), the `<head>` node ([`document.head`](https://developer.mozilla.org/en-US/docs/Web/API/Document/head)), and even the doctype ([`document.doctype`](https://developer.mozilla.org/en-US/docs/Web/API/Document/doctype)).
+
+![A screenshot of the Chrome debugger console displaying those properties](first_document_properties.png)
+
+### Querying Elements
```
evelynhathaway: I think it may be a good idea to bring up `document.getElement(s)By___()` because `document.getElementById()`:
- is frequently used
- has better browser support than querying
- and is more performant than `document.querySelector("#unique-id")`

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+const mainTextElement = document.querySelector('#mainText');
+```
+
+> The `#` in the `#mainText` is the CSS selector syntax for selecting an element based on it's `id`. If you had a CSS selector of `#testing`, you'd be looking for an element with the following attribute value:
+>
+> ```
+> id="testing"
+> ```
+
+This method will return a reference to the element as rendered in the DOM. [While we'll be covering more of what this reference is able to do later](#element-class), we can do a quick bit of code to show that it's the real element we intended to query:
```
evelynhathaway:
```suggestion
This method will return a reference to the element as rendered in the DOM. [While we'll be covering more of what this reference is able to do later](#element-class), but for now we can execute this quick bit of code to show that it's the element we intended to query:
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+```javascript
+console.log(mainTextElement.innerHTML); // This will output the HTML that we used to write this element
+```
+
+![A screenshot of the Chrome debugger running the above code](query_selector.png)
+
+We also have the ability to gain a reference to many elements at once. GIven the same HTML document as before, let's say we want to see how many elements have the `bolded` class applied to it. We're able to do so using [the `document` `querySelectorAll` method](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll).
```
evelynhathaway:
```suggestion
We also have the ability to gain a reference to many elements at once. Given the same HTML document as before, let's say we want to see how many elements have the `bolded` class applied to it. We're able to do so using [the `querySelectorAll` method on the `document`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll).
```
- For the reference to the method name, I would either make it consistent with the mention to the other method, or I would make it plainly `document.querySelectorAll`.

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+![A screenshot of Chrome running the above code](query_selector_all.png)
+
+> It's worth mentioning that the way `querySelector` works is not the same [way that the browser checks a node against the CSS selector data when the browser "visits" that node](#how-the-browser-uses-the-dom). `querySelector` and `querySelectorAll` work from a more top-down perspective where it searches the elements one-by-one against the query. First, it finds the top-most layer of the CSS selector. Then it will move to the next item and so-on-so forth until it returns the expected results.
+
+## Element Base Class {#element-class}
+
+While `innerHTML` has been used to demonstrate that the element that's gathered is in fact the element that was queried, there are many _many_ more properties and methods that can be ran on an element reference.
+
+When an element is queried and returned, you're given a reference to that element through the [`Element` base class ](https://developer.mozilla.org/en-US/docs/Web/API/Element). This class is what contains the properties and methods that you can use to access and modify metadata about the element with.
```
evelynhathaway:
```suggestion
When an element is queried and returned, you're given a reference to that element through the [`Element` base class](https://developer.mozilla.org/en-US/docs/Web/API/Element). This class is what contains the properties and methods that you can use to access and modify metadata about the element with.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+For example, let's say that I wanted to see the width and height an element has when rendered on screen. [Using the `Element.prototype.getBoundingClientRect` method](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect), you can get all of that information and more:
+
+```javascript
+const mainTextElement = document.querySelector('#mainText');
+console.log(mainTextElement.getBoundingClientRect());
+// Will output: DOMRect {x: 8, y: 16, width: 638, height: 18, top: 16, …}
+```
+
+> While the explaination behind the `Element.prototype` is a lengthy one (an article on it's own to be sure), suffice it to say that there's a base class for all element references found using `querySelector`. This base class contains a myriad of methods and properties. The `.prototype` loosely refers to those properties and methods in question.
```
evelynhathaway:
```suggestion
> While the explanation behind the `Element.prototype` is a lengthy one (an article on its own to be sure), suffice it to say that there's a base class for all element references found using `querySelector`. This base class contains a myriad of methods and properties. The `.prototype` loosely refers to those properties and methods in question.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+```html
+<div id="divToList">
+	<div>Bananas</div>
+	<div>Apples</div>
+	<div>Oranges</div>
+</div>
+```
+
+We could update this list to include the `role`s and `aria-label`s in order to make this non-symantic HTML more relevant with how it reflects it's metadata to the browser.
```
evelynhathaway:
```suggestion
We could update this list to include the `role`s and `aria-label`s in order to make this non-semantic HTML more relevant with how it reflects its metadata to the browser.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+This metadata that we place directly on the elements themselves are called `attributes` and are part of the HTML specification (also refered to as the HTML API in this document). This metadata can be accessed and modified from JavaScript by using the `Element`'s `getAttribute` to read the key-value pairing and `setAttribute` to set the value to that attribute on an element.
```
evelynhathaway:
```suggestion
This metadata that we place directly on the elements themselves are called `attributes` and are part of the HTML specification (also referred to as the HTML API in this document). This metadata can be accessed and modified from JavaScript by using the `Element`'s `getAttribute` to read the key-value pairing and `setAttribute` to set the value to that attribute on an element.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+```html
+<div id="divToList" role="list" aria-label="My favorite fruits">
+	<div role="listitem">Bananas</div>
+	<div role="listitem">Apples</div>
+	<div role="listitem">Oranges</div>
+</div>
+```
+
+Which is significantly more accessible for users that utilize screen readers, [as mentioned previously](#accessibility). You'll notice that despite not having any of the ARIA attributes prior, the `setAttribute` was able to implicitly create them with the newly placed values.
```
evelynhathaway:
```suggestion
... which is significantly more accessible for users that utilize screen readers, [as mentioned previously](#accessibility). You'll notice that despite not having any of the ARIA attributes prior, the `setAttribute` was able to implicitly create them with the newly placed values.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
+---
+{
+	title: 'Understanding The DOM: How Browsers Show Content on Screen',
```
evelynhathaway:
```suggestion
	title: 'Understanding The DOM: How Browsers Show Content On-Screen',
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+_The browser takes the items defined in the HTML and turns them into a tree that the browser understands how to lay out and draw on the screen_. That tree, internally, might look something like this:
+
+![A chart showing the document object model layout of the above code. It shows that the 'main' tag is the parent to a 'ul' tag, and so on](./dom_tree.svg "Diagram showing the above code as a graph")
+
+> This is an oversimplified example of how the browser interprets HTML, but gets the job done to convey introductory information.
+
+Let's see how this is done.
+
+At the root of any HTML file, you have three things: Tags, attributes, and text content.
```
evelynhathaway:
```suggestion
At the root of any HTML file, you have three things: tags, attributes, and text content.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+```javascript
+element.dataset.userInfo = {name: "Tony"}
+console.log(element.dataset.userInfo) // "[object Object]"
+/**
+ * "[object Object]" is because it's running `Object.prototype.toString()`
+ * to convert the object to a string to store on the attribute
+ */
+```
+
+> If you're having a difficult time understanding why `toString` is bring ran or what `prototype` is doing here, don't worry: You're in good company. The JavaScript prototype system is complex and can be difficult to follow.
```
evelynhathaway:
```suggestion
> If you're having a difficult time understanding why `toString` is bring ran or what `prototype` is doing here, don't worry; you're in good company. The JavaScript prototype system is complex and can be difficult to follow.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+While attributes can be of great use to store data about an element, there's a limitation: Values are always stored as strings. This means that objects, arrays, and other non-string primitives must find a way to go to and from strings when being read and written.
+
+> While you've seen `style` attribute be read and written to by an object interface, if you inspect the element or use the `getAttribute` to access the attribute's HTML API value, you'll find that it's really a string with a pleasant API wrapped around it that lets you use an object to interface with the attribute value.
+>
+> ```javascript
+> console.log(mainTextElement.getAttribute('style')); // This will return a string value, despite the API that lets you use an object to read and write
+> ```
+>
+> The reasoning behind this incongruity is due to [the implicit mapping of the HTML API and the `Element` API as mentioned at the start of the previous section](#element-properties). The limitations described here will also apply to the HTML API of those types of properties.
```
evelynhathaway:
```suggestion
> The reasoning behind this incongruity is due to [the implicit mapping of the HTML API and the `Element` API, as mentioned at the start of the previous section](#element-properties). The limitations described here will also apply to the HTML API of those types of properties.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+// index.js
+const listEl = document.querySelector('#list');
+console.log(listEl.dataset.listitems); // '2'
+listEl.dataset.listitems = 3;
+console.log(listEl.dataset.listitems); // '3'
+```
+
+![Demonstrating that dataset values are able to be read and written](list_dataset.png)
+
+If you'll notice, I wrote the string `'3'` instead of the numerical value `3` in the code sample's outputs in the comments despite using the numerical `3` to set the value. This behavior is due to how default non-string values are saved to attributes.
```
evelynhathaway:
```suggestion
Note that I wrote the string `'3'` instead of the numerical value `3` in the code sample's outputs in the comments despite using the numerical `3` to set the value. This behavior is due to how default non-string values are saved to attributes.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+console.log(element.dataset.userInfo) // "[object Object]"
+/**
+ * "[object Object]" is because it's running `Object.prototype.toString()`
+ * to convert the object to a string to store on the attribute
+ */
+```
+
+> If you're having a difficult time understanding why `toString` is bring ran or what `prototype` is doing here, don't worry: You're in good company. The JavaScript prototype system is complex and can be difficult to follow.
+>
+> For now, it will suffice to just know that you're only able to store strings in element attribute.
```
evelynhathaway:
```suggestion
> For now, it will suffice just to know that you're only able to store strings in an element attribute.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+But nothing from the green square's `eventListener`.
+
+<iframe src="https://stackblitz.com/edit/event-capture-demo?ctl=1&embed=1&file=index.js&hideExplorer=1&hideNavigation=1" style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
+
+You'll also notice that if you click on the green square, you'll never see the `"A click handled on green using capture"` message. This is due to the `stopPropagation` as mentioned before. The click is being registered on the red square first and then stopped on the blue square.
+
+# Conclusion
+
+This post has been filled to the brim with information. 😵 Even I, the author, have had to have a few amazing folks take a re-read to confirm what I've written. Please don't be afraid or ashamed to re-read anything that might not have made sense or to revisit the post whenever a question arises. Hopefully this has been a helpful exploration of the DOM and the ways you interact with it using code.
```
evelynhathaway:
```suggestion
This post is filled to the brim with information. 😵 Even I, the author, had a few amazing folks give it a re-read to confirm what I've written. Please don't be afraid or ashamed to re-read anything that might not have made sense or to revisit the post whenever a question arises. Hopefully, this has been a helpful exploration of the DOM and the ways you interact with it using code.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+Please ask any questions or comments in our comments section and remember that we have [a Discord](https://discord.gg/FMcvc6T) for further conversation including questions!
```
evelynhathaway:
```suggestion
Please ask any questions or comments in our comments section and remember that we have [a Discord](https://discord.gg/FMcvc6T) for further conversation, including any questions!
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+> If you're having a difficult time understanding why `toString` is bring ran or what `prototype` is doing here, don't worry: You're in good company. The JavaScript prototype system is complex and can be difficult to follow.
+>
+> For now, it will suffice to just know that you're only able to store strings in element attribute.
+
+
+
+## Events {#events}
+
+Just as your browser uses the DOM to know where to place elements to display for the user and metadata about the elements in order to interact with screenreaders as-expected, your browser also utilizes the DOM for knowing how to handle user interactions. The way your browser handles user interaction is by listening for _events_ that occur when the user takes action or other noteworthy changes occur.
```
evelynhathaway: This is extremely wordy and should be split into multiple sentences if everything is needed.

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+For example, when you have a form that includes a default button, when that button is pressed it will fire a `submit` event that will then _bubble_ up the DOM tree until it finds a [`<form>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form). This element will by default then run a [`GET` HTML request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) to the server once it recieves a `submit` event.
```
evelynhathaway:
```suggestion
For example, say you have a form that includes a default `<button>` element. When that button is pressed, it fires a `submit` event that then _bubbles_ up the DOM tree until it finds a [`<form>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form). By default, this `<form>` element then sends a [`GET` HTML request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) to the server once it receives the `submit` event.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+![The bubble flow of the `submit` event](./submit_form.svg)
+
+_Bubbling_, as shown here, is the default behavior of any given event. It reflects it's behavior of a given event moving up the DOM tree to the nodes above it as parents. Parent nodes can respond to these events as expected, stop their upward motion on the tree, and more.
```
evelynhathaway:
```suggestion
_Bubbling_, as shown here, is the default behavior of all events. It reflects the behavior of the given event, moving up the DOM tree to the parent nodes above the element the event originated from. Parent nodes can respond to these events as expected, stop their upward motion on the tree, and more.
```
- This is my best guess on how to make this paragraph more readable. The second sentence, in particular, wasn't clear to me.

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+		greenEl.addEventListener('click', () => {
+			console.log("A click handled on green using bubbling");
+		}, false);
+	</script>
+</body>
+</html>
+```
+
+
+In this example, we're adding click listeners to three squares, each one smaller then their parent square. This allows us to see in our console the effect of bubbling. If you click on the red square, you'd expect the event to bubble up to `<body>`, but not down to `#green`. Likewise, if you clicked on the green square, you'd expect the event to bubble up to both `#blue` and `#red` as well as `<body>`.
```
evelynhathaway:
```suggestion
In this example, we're adding click listeners to three squares, each one smaller than their parent square. This allows us to see the effect of bubbling in our console. If you click on the red square, you'd expect the event to bubble up to `<body>`, but not down to `#green`. Likewise, if you clicked on the green square, you'd expect the event to bubble up to both `#blue` and `#red` as well as `<body>`.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+![The event bubbles upwards from green to blue but then is stopped by the stopPropagate call](./stop_propagration.svg)
+
+You can see a running example of this here:
+
+<iframe src="https://stackblitz.com/edit/event-bubbling-demo?ctl=1&embed=1&file=index.js&hideExplorer=1&hideNavigation=1" style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
+
+### Capturing {#event-capturing}
+
+Bubbling isn't the only way events are able to move; Just as they can move up from the bottom, they can also move from the top down. This method of emitting events is known as _capture mode_.
```
evelynhathaway:
```suggestion
Bubbling isn't the only way events are able to move. Just as they can move up from the bottom, they can also move from the top down. This method of emitting events is known as _capture mode_.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+```
+"A click handled on red using capturing"
+"A click handled on blue using capturing"
+```
+
+But nothing from the green square's `eventListener`.
+
+<iframe src="https://stackblitz.com/edit/event-capture-demo?ctl=1&embed=1&file=index.js&hideExplorer=1&hideNavigation=1" style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
+
+You'll also notice that if you click on the green square, you'll never see the `"A click handled on green using capture"` message. This is due to the `stopPropagation` as mentioned before. The click is being registered on the red square first and then stopped on the blue square.
```
evelynhathaway:
```suggestion
You'll also notice that if you click on the green square, you'll never see the `"A click handled on green using capture"` message. This is due to the `stopPropagation`, as mentioned before. The click is being registered on the red square first and then stopped on the blue square.
```

content/blog/understanding-the-dom/index.md
crutchcorn: I don't generally disagree with any of these points, but I think this post is meant to be more focused on concept - not implementation

content/blog/understanding-the-dom/index.md
crutchcorn:
```suggestion
Just as your browser uses the DOM to handle on-screen content visibility, your browser also utilizes the DOM for knowing how to handle user interactions. The way your browser handles user interaction is by listening for _events_ that occur when the user takes action or other noteworthy changes occur.
```

content/blog/understanding-the-dom/index.md
crutchcorn:
```suggestion
_Bubbling_, as shown here, is the default behavior of any given event. Its behavior is to move an event up the DOM tree to the nodes above it, moving from child to parent until it hits the root. Parent nodes can respond to these events as expected, stop their upward motion on the tree, and more.
```

I'd like to leave the first sentence as-is

content/blog/understanding-the-dom/index.md
evelynhathaway: Ahhh I gotcha

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+	description: 'Learn how the browser internally handles HTML and CSS to show the user webpages on-screen',
+	published: '2019-11-26T22:12:03.284Z',
+	authors: ['crutchcorn'],
+	tags: ['dom', 'browser internals'],
+	attached: [],
+	license: 'cc-by-nc-sa-4'
+}
+---
+
+Any web application relies on some fundamental technologies: HTML, CSS, and JavaScript. Even advanced front-end JavaScript frameworks such as Angular, React, or Vue will utilize some level of HTML to load the JavaScript. That said, how the browser handles HTML and CSS under-the-hood can be quite the mystery. In this article, I'm going to explain what the browser does to understand what it should show to the user.
```
adueppen:
```suggestion
Every web application relies on some fundamental technologies: HTML, CSS, and JavaScript. Even advanced front-end JavaScript frameworks such as Angular, React, or Vue still utilize some amount of HTML to load the JavaScript. That said, how the browser handles HTML and CSS under-the-hood can be quite the mystery. In this article, I'm going to explain what the browser does to understand what it should show to the user.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+# The DOM {#the-dom}
+
+Just as the source code of JavaScript programs are broken down to abstractions that are more easily understood by the computer, so too is HTML. HTML, initially being derived from [SGML (the basis for XML as well)](https://en.wikipedia.org/wiki/Standard_Generalized_Markup_Language), actually _forms a tree structure in memory_ in order to [describe the relationships, layout, and executable tasks for items in the tree](#how-the-browser-uses-the-dom). This tree structure in memory is _called the Document Object Model_ (or _DOM_ for short).
```
adueppen:
```suggestion
Just as the source code of JavaScript programs are broken down to abstractions that are more easily understood by the computer, so too is HTML. HTML, initially being derived from [SGML (the basis for XML as well)](https://en.wikipedia.org/wiki/Standard_Generalized_Markup_Language), _forms a tree structure in memory_ in order to [describe the relationships, layout, and executable tasks for items in the tree](#how-the-browser-uses-the-dom). This tree structure in memory is _called the Document Object Model_ (or _DOM_ for short).
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+<header>
+  <!-- An "a" tag with an "href" attribute -->
+  <a href="example.com">
+    <!-- A text node -->
+    Example Site
+  </a>
+</header>
+```
+
+When you type a tag, like `<header>` or `<a>`, you're creating an _element node_. These nodes then compose to create _"leaves"_ on the DOM tree. Attributes are then able to manually add information to said nodes. When you have another element node inside of a separate element node, you add a _"child"_ to said node. This relationship between the nodes allows you to preserve metadata between them and allows CSS to apply.
```
adueppen:
```suggestion
When you type a tag, like `<header>` or `<a>`, you're creating an _element node_. These nodes are then composed to create _"leaves"_ on the DOM tree. Attributes are then able to manually add information to these nodes. When you have one element node inside of a separate one, you add a _"child"_ to said node. The relationship between the nodes allows metadata, CSS properties, and more to be preserved.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+There's also the idea of a _"sibling"_ node. When a node's parent has more than one child, those other nodes are that child node's _"siblings"_.
+
+![A chart showing the relationships between parents and siblings](./dom_relationships.svg)
+
+Altogether, the terminologies used to refer between these node "leaves" are extremely similar to the terminology often used with family trees.
```
adueppen:
```suggestion
Altogether, the terminology used to refer to the nodes and their various relationships is extremely similar to the terminology often used with family trees.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+There are some rules to this tree that's created from these "nodes":
```
adueppen:
```suggestion
There are some rules for the tree that's created from these nodes:
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+- There must be one "root" or "trunk" node, and there cannot be more than one root
+- There must be a one-to-many relationship with parents and children
+	- A node may have many children
+	- But may not have many parents
```
adueppen: replace with
```
- There must be a one-to-many relationship with parents and children. A node:
	- May have many children
	- Cannot have more than one parent
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+- A non-root node may have many siblings as a result of the parent having many children
+
+![A chart showing the aforementioned rules of the node relationships](./dom_relationship_rules.svg)
+
+### How It's Used By The Browser {#how-the-browser-uses-the-dom}
+
+This tree tells the browser all of the information the browser needs to execute tasks in order to display and handle interaction with the user. For example, when the following CSS is applied to an `index.html` file:
```
adueppen:
```suggestion
This tree tells the browser all of the information it needs to execute tasks in order to display and handle interaction with the user. For example, when the following CSS is applied to this HTML file:
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+<main id="a">
+	<p id="e"></p>
+  <ul id="b">
+    <li id="c"></li>
+    <li id="d"></li>
+  </ul>
+</main>
+```
+
+The browser is able to keep in mind, while it's moving through the tree, that it needs to find an element with the `ID` of `b` and then mark it's `<li>` children with a red background. They're "children" because the DOM tree keeps that relationship info that's defined by the HTML.
```
adueppen:
```suggestion
While moving through the tree, the browser engine can keep track of the fact that it needs to find an element with the `ID` of `b` and then mark its `<li>` children with a red background. They're "children" because the DOM tree preserves the relationship defined by the HTML.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+![A chart showing the 'ul' tag highlighted in green with the children 'li' tags marked in red](./dom_tree_with_css.svg "Diagram showing the above code as a graph")
+
+> The `<ul>` element is marked as green just to showcase that it is the element being marked by the first part of the selector.
+
+Typically, the browser will "visit" it's nodes in a specific order. For example, in the above chart, the browser might start at the `<main>` tag, then go to the `<p>` tag, then visit the `<ul>` tag, and finally the two children in order from left-to-right (`<li id="c">` , `<li id="d">`).
+
+The browser, knowing what CSS to look for, is able to see the `<ul>` with the correct ID and know to mark its children with the correct metadata that matches the selector with the relevant CSS.
+
+This tree relationship also enables CSS selectors such as the [general sibling selector (`~`)](https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator) or the [adjacent sibling selector (`+`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator) to find siblings to a given selector.
```
adueppen:
```suggestion
This tree relationship also enables more sophisticated CSS selectors such as [general sibling selector (`~` )](https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator) or the [adjacent sibling selector (`+`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator) to find siblings of a given selector.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+![A showcase of the above selectors and how they always look forward, never behind](css_selectors_demo.svg)
+
+> Interestingly, one of the questions that I've often heard asked concerns a "parent selector". The idea behind the question is that the [direct child selector (`>`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator) exists, so why not have the ability to mark any parent of `.classname` selectors?
```
adueppen:
```suggestion
> Interestingly, one of the questions that I've often heard asked concerns a "parent selector". The idea behind the question is that the [direct child selector (`>`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator) exists, so why not have the ability to select any parent of `.classname` selectors?
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+>
+> The answer behind that: performance. The [W3 organization](https://www.w3.org/Style/CSS/#specs) (the organization who maintains the HTML and CSS standard specification) points to the tree structure of the DOM and the algorithm behind how browsers traverse the DOM (or, "visit" the nodes in order to figure out what CSS to apply)  as not being performant when allowing parent selectors.
```
adueppen:
```suggestion
> The answer behind that? Performance. The [W3 Consortium](https://www.w3.org/Style/CSS/#specs) (the organization that maintains the HTML and CSS standard specifications) points to the tree structure of the DOM and the algorithm used to traverse the DOM (visiting the nodes in order to figure out what CSS to apply) as not being performant when allowing parent selectors.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+>
+> The reason behind this is that browsers read from top-down in the DOM and apply CSS as they find matching nodes; CSS doesn't command the browser to do anything to the DOM, but rather provides the metadata for the DOM to apply the relevant CSS when the browser comes across that specific node.
```
adueppen:
```suggestion
> This happens because browsers read from top-to-bottom in the DOM and apply CSS as they find matching nodes; CSS doesn't command the browser to do anything to the DOM, but rather provides the metadata for the DOM to apply the relevant CSS when the browser comes across that specific node.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+>
+> As mentioned before, they start at the root node, keep notes on what they've seen, then move to children. Then, they move to siblings, etc. Specific browsers may have slight deviations on this algorithm, but for the most part, they don't allow for upwards vertical movement of nodes within the DOM.
+
+
+
+# Using The Correct Tags {#accessibility}
+
+HTML, as a specification, has tons of tags that can be used at one's disposal. These tags contain various metadata internally to the browser to provide information about what you want to render in the DOM. This metadata can then be handled by the browser how it sees fit; it may apply default CSS styling, it may change the default interaction the user has with it, or even what behavior that element has upon clicking on it (in the case of a button in a form).
```
adueppen:
```suggestion
HTML, as a specification, has tons of tags that are able to be used at one's disposal. These tags contain various pieces of metadata to provide information to the browser about how they should be rendered in the DOM. This metadata can then be handled by the browser how it sees fit; it may apply default CSS styling, it may change the default interaction the user has with it, or even what behavior that element has upon clicking on it (in the case of a button in a form).
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+Some of these tag defaults are part of the specification, while others are left up to the browser vendor to decide. This is why, in many instances, developers may choose to use something like [`normalize.css`](https://github.com/necolas/normalize.css/) to set all of the element CSS defaults to an explicit set of defaults. Doing so can avoid having the UI of a webpage look different from browser-to-browser thanks to deviations on default CSS styling on specific tags.
```
adueppen:
```suggestion
Some of these tag defaults are part of the specification, while others are left up to the browser vendor to decide. This is why, in many instances, developers may choose to use something like [`normalize.css`](https://github.com/necolas/normalize.css/) to set all of the element CSS defaults to an explicit set of defaults. Doing so can avoid having the UI of a webpage look different from browser to browser thanks to deviations on default CSS styling on specific tags.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+<ol>
+	<li>Bananas</li>
+	<li>Apples</li>
+	<li>Oranges</li>
+</ol>
+```
+
+In this example, both the browsers as well as Google's scraper bots are able to discern that this is a list with three list items within it.
+
+> While there ARE tags that may potentially  impact SEO somewhat significantly, it's unlikely `<ul>` and `<li>` would significantly impact your SEO scores.
```
adueppen:
```suggestion
> While there ARE tags that may potentially impact SEO somewhat significantly, it's unlikely `<ul>` and `<li>` would significantly impact your SEO scores.
```

adueppen: (duplicate space)

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+>
+> Needless to say, it's still good to use semantic (correctly tagged) HTML as people that use screen-readers and other assistive technologies benefit greatly from these minor changes. Additionally, it can make code more readable and parsable with automated tools.
```
adueppen:
```suggestion
> Needless to say, it's still good to use semantic (correctly tagged) HTML as people that use screen-readers and other assistive technologies benefit greatly from these minor changes. Additionally, it can make it easier for automated tools to read and parse your code.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+We're able to even add further metadata to an element by using attributes. For example, let's say that I want to add a title to the list to be read upon a screen reader gaining focus on the element; we could use the `aria-label` attribute:
```
adueppen:
```suggestion
We can add even more metadata to an element by using attributes. For example, let's say that I want to add a title to the list to be read when a screen reader gains focus on the element. This can be solved by using the `aria-label` attribute:
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+```html
+<ol aria-label="My favorite fruits">
+	<li>Bananas</li>
+	<li>Apples</li>
+	<li>Oranges</li>
+</ol>
+```
+
+In fact, the default metadata that is defaulted by specific tags can be directly applied manually to an element of a different tag. The metadata that is passed to the browser when using `<li>` is typically involving that element pertaining to a `listitem`, using the `role` attribute, we can add that information to a `<div>` itself.
```
adueppen:
```suggestion
In fact, the metadata that specific tags have by default can be manually applied to an element of a different type. The metadata that is passed to the browser when using `<li>` is typically involving that element pertaining to a `listitem`, using the `role` attribute, we can add that information to a `<div>` itself.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+Let's look at some of the built-in utilities at our disposal for doing so:
+
+- [The `document` global object](document-global-object)
+- [The `Element` base class](element-class)
+- [The event system](#events)
+
+## Document Global Object {#document-global-object}
+
+[As mentioned before, the DOM tree must contain one root node](#the-dom). This node, for any instance of the DOM, is the document entry point. When in the browser, this entry point is exposed to the developer with [the global object `document`](https://developer.mozilla.org/en-US/docs/Web/API/Document). This object has various methods and properties to assist in a meaningful way. For example, given a standard HTML5 document:
```
adueppen:
```suggestion
[As mentioned before, the DOM tree must contain one root node](#the-dom). This node, for any instance of the DOM, is the document entry point. When in the browser, this entry point is exposed to the developer with [the global object `document`](https://developer.mozilla.org/en-US/docs/Web/API/Document). This object has various helpful methods and properties. For example, given a standard HTML5 document:
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+</html>
+```
+
+The `document` object has the ability to get the `<body>` node ([`document.body`](https://developer.mozilla.org/en-US/docs/Web/API/Document/body)), the `<head>` node ([`document.head`](https://developer.mozilla.org/en-US/docs/Web/API/Document/head)), and even the doctype ([`document.doctype`](https://developer.mozilla.org/en-US/docs/Web/API/Document/doctype)).
+
+![A screenshot of the Chrome debugger console displaying those properties](first_document_properties.png)
+
+### Querying Elements
+
+Besides containing static references to `<body>` and `<head>`, there is also a way to query for any element by using CSS selectors. For example, if we wanted to get a reference to the single element with the `id` of `mainText`, we could use the CSS selector for an id, combined with [the `querySelector` method on the `document`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector):
```
adueppen:
```suggestion
Besides containing static references to `<body>` and `<head>`, there is also a way to query any element by using CSS selectors. For example, if we wanted to get a reference to the single element with the `id` of `mainText`, we could use the CSS selector for an id, combined with [the `querySelector` method on the `document`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector):
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+const mainTextElement = document.querySelector('#mainText');
+```
+
+> The `#` in the `#mainText` is the CSS selector syntax for selecting an element based on its `id`. If you had a CSS selector of `#testing`, you'd be looking for an element with the following attribute value:
+>
+> ```
+> id="testing"
+> ```
+
+This method will return a reference to the element as rendered in the DOM. [While we'll be covering more of what this reference is able to do later](#element-class), but for now we can execute this quick bit of code to show that it's the element we intended to query:
```
adueppen:
```suggestion
This method will return a reference to the element as rendered in the DOM. [While we'll be covering more of what this reference is able to do later](#element-class), for now we can execute this quick bit of code to show that it's the element we intended to query:
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+console.log(boldedElements[0].innerHTML); // Will output the HTML for that element
+```
+
+![A screenshot of Chrome running the above code](query_selector_all.png)
+
+> It's worth mentioning that the way `querySelector` works is not the same [way that the browser checks a node against the CSS selector data when the browser "visits" that node](#how-the-browser-uses-the-dom). `querySelector` and `querySelectorAll` work from a more top-down perspective where it searches the elements one-by-one against the query. First, it finds the top-most layer of the CSS selector. Then it will move to the next item and so-on-so forth until it returns the expected results.
+
+## Element Base Class {#element-class}
+
+While `innerHTML` has been used to demonstrate that the element that's gathered is in fact the element that was queried, there are many _many_ more properties and methods that can be ran on an element reference.
```
adueppen:
```suggestion
While `innerHTML` has been used to demonstrate that the element that's gathered is in fact the element that was queried, there are many _many_ more properties and methods that can be run on an element reference.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+When an element is queried and returned, you're given a reference to that element through the [`Element` base class](https://developer.mozilla.org/en-US/docs/Web/API/Element). This class is what contains the properties and methods that you can use to access and modify metadata about the element with.
```
adueppen:
```suggestion
When an element is queried and returned, you're given a reference to that element through the [`Element` base class](https://developer.mozilla.org/en-US/docs/Web/API/Element). This class is what contains the properties and methods that you can use to access and modify the element's metadata.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+// Will output: DOMRect {x: 8, y: 16, width: 638, height: 18, top: 16, …}
+```
+
+> While the explanation behind the `Element.prototype` is a lengthy one (an article on its own to be sure), suffice it to say that there's a base class for all element references found using `querySelector`. This base class contains a myriad of methods and properties. The `.prototype` loosely refers to those properties and methods in question.
+>
+> This means that all queried elements will have their own `getBoundingClientRect` methods.
+
+### Attributes {#html-attributes}
+
+[As covered before in this post, elements are able to have _attributes_ that will apply metadata to an element for the browser to utilize.](#accessibility) However, what I may not have mentioned is that you're able to read, write, and modify that metadata using JavaScript.
```
adueppen:
```suggestion
[As covered earlier, elements are able to have _attributes_ that will apply metadata to an element for the browser to utilize.](#accessibility) However, what I may not have mentioned is that you're able to read and write that metadata, as well as applying new metadata, using JavaScript.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+```html
+<div id="divToList">
+	<div>Bananas</div>
+	<div>Apples</div>
+	<div>Oranges</div>
+</div>
+```
+
+We could update this list to include the `role`s and `aria-label`s in order to make this non-semantic HTML more relevant with how it reflects its metadata to the browser.
```
adueppen:
```suggestion
We could update this list to include the `role`s and `aria-label`s in order to make this non-semantic HTML more relevant in terms of how it reflects its metadata to the browser.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+This metadata that we place directly on the elements themselves are called `attributes` and are part of the HTML specification (also referred to as the HTML API in this document). This metadata can be accessed and modified from JavaScript by using the `Element`'s `getAttribute` to read the key-value pairing and `setAttribute` to set the value to that attribute on an element.
```
adueppen:
```suggestion
This metadata that we place directly on the elements themselves are called `attributes` and are part of the HTML specification (also referred to as the HTML API in this document). This metadata can be accessed and modified from JavaScript by using the `Element`'s [`getAttribute`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute) to read the key-value pairing and [`setAttribute`](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute) to set the value to that attribute on an element.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+Let's look at how we can set the `role` and `aria-label`s in the DOM using JavaScript:
+
+```javascript
+const divToListEl = document.querySelector('#divToList');
+// Get the `role` attribute to demonstrate that there's no currently present role
+console.log(divToList.getAttribute('role')); // `null`
```
adueppen:
```suggestion
console.log(divToListEl.getAttribute('role')); // `null`
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+// Let's set a role that emulates a `list`
+// Set the value from the HTML API using the Element method `setAttribute`
+divToListEl.setAttribute('role', 'list');
+// And let's add an aria-label, for good measure
+divToListEl.setAttribute('aria-label', 'My favorite fruits');
+// Get the value from the HTML API using the Element method `getAttribute`
+console.log(divToList.getAttribute('role')); // `'list'`
```
adueppen:
```suggestion
console.log(divToListEl.getAttribute('role')); // `'list'`
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+// Using the CSS selector to get the children of the divs
+const listItems = document.querySelectorAll('#divToList > *');
+
+// Now, for all of the items in that list, let's use an aria `role` to make them reflect as listitems in their metadata to the browser
+for (var i = 0; i < listItems.length; i++) {
```
adueppen:
```suggestion
for (let i = 0; i < listItems.length; i++) {
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+	listItems[i].setAttribute('role', 'listitem');
+}
+```
+
+Once this is all ran, if you inspect the elements tab in your debugger, you should be left with HTML that looks like this:
```
adueppen:
```suggestion
Once this is run, if you inspect the elements tab in your debugger, you should be left with HTML that looks like this:
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+```
+
+![Demonstrating that dataset values are able to be read and written](list_dataset.png)
+
+Note that I wrote the string `'3'` instead of the numerical value `3` in the code sample's outputs in the comments despite using the numerical `3` to set the value. This behavior is due to how default non-string values are saved to attributes.
+
+By default, the primitive's `toString` will be called to store values.
+
+```javascript
+element.dataset.userInfo = {name: "Tony"}
```
adueppen:
```suggestion
element.dataset.userInfo = {name: "Tony"};
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+console.log(element.dataset.userInfo) // "[object Object]"
```
adueppen:
```suggestion
console.log(element.dataset.userInfo); // "[object Object]"
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+/**
+ * "[object Object]" is because it's running `Object.prototype.toString()`
+ * to convert the object to a string to store on the attribute
+ */
+```
+
+> If you're having a difficult time understanding why `toString` is bring ran or what `prototype` is doing here, don't worry; you're in good company. The JavaScript prototype system is complex and can be difficult to follow.
```
adueppen:
```suggestion
> If you're having a difficult time understanding why `toString` is bring run or what `prototype` is doing here, don't worry; you're in good company. The JavaScript prototype system is complex and can be difficult to follow.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+>
+> For now, it will suffice just to know that you're only able to store strings in an element attribute.
+
+
+
+## Events {#events}
+
+Just as your browser uses the DOM to handle on-screen content visibility, your browser also utilizes the DOM for knowing how to handle user interactions. The way your browser handles user interaction is by listening for _events_ that occur when the user takes action or other noteworthy changes occur.
```
adueppen:
```suggestion
Just as your browser uses the DOM to handle on-screen content visibility, your browser also utilizes the DOM for knowing how to handle user interactions. The way your browser handles user interaction is by listening for _events_ that occur when the user takes action or when other noteworthy changes occur.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+
+For example, say you have a form that includes a default `<button>` element. When that button is pressed, it fires a `submit` event that then _bubbles_ up the DOM tree until it finds a [`<form>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form). By default, this `<form>` element then sends a [`GET` HTML request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) to the server once it receives the `submit` event.
```
adueppen:
```suggestion
For example, say you have a form that includes a default `<button>` element. When that button is pressed, it fires a `submit` event that then _bubbles_ up the DOM tree until it finds a [`<form>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form). By default, this `<form>` element sends a [`GET` HTML request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) to the server once it receives the `submit` event.
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+		const blueEl = document.querySelector('#blue');
+		const greenEl = document.querySelector('#green');
+
+		redEl.addEventListener('click', () => {
+			console.log("A click handled on red using bubbling");
+			// This is set to false in order to use bubbling. We'll cover the `true` case later on
+		}, false);
+
+		blueEl.addEventListener('click', (event) => {
+			// Stop the click event from moving further up in the bubble
```
adueppen:
```suggestion
			// Stop the click event from moving further up in the DOM tree ("bubbling up")
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+  console.log("A click handled on blue using capturing");
+}, true);
+
+
+greenEl.addEventListener('click', () => {
+  console.log("A click handled on green using capturing");
+}, true);
+```
+
+As demonstrated by the code above, `stopPropagation` works as you might expect it to as well!
```
adueppen:
```suggestion
As demonstrated by the code above, `stopPropagation` works as you might expect it to in capture mode as well!
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+![stopPropagation works similarly to how it does in bubble mode, just that it stops events from moving _down_ the tree](./capture_stop_propagation.svg)
+
+This means that when the user clicks on the red square, you'll see the following in your console:
+
+```
+"A click handled on red using capturing"
+"A click handled on blue using capturing"
+```
+
+But nothing from the green square's `eventListener`.
```
adueppen:
```suggestion
You won't see anything from the green square's `eventListener`, however.
```

adueppen:
I did also notice that "screen reader" and "screen-reader" were used inconsistently in the post. Although either one would work fine, only one should be used ideally.

content/blog/understanding-the-dom/index.md
crutchcorn: Closing, as this is just a shift of tone and not really a required change per grammar/spelling

content/blog/understanding-the-dom/index.md
crutchcorn: Again, this is just a tonal change by removing a word. Closing as a result

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+There's also the idea of a _"sibling"_ node. When a node's parent has more than one child, those other nodes are that child node's _"siblings"_.
+
+![A chart showing the relationships between parents and siblings](./dom_relationships.svg)
+
+Altogether, the terminologies used to refer between these node "leaves" are extremely similar to the terminology often used with family trees.
+
+There are some rules to this tree that's created from these "nodes":
+
+- There must be one "root" or "trunk" node, and there cannot be more than one root
+- There must be a one-to-many relationship with parents and children
```
crutchcorn:
```suggestion
- There must be a one-to-many relationship with parents and children. A node:
```

content/blog/understanding-the-dom/index.md
```diff
@@ -0,0 +1,519 @@
...
+	- A node may have many children
```
crutchcorn:
```suggestion
	- May have many children
```

content/blog/understanding-the-dom/index.md
crutchcorn:
```suggestion
	- Cannot have more than one parent
```

content/blog/understanding-the-dom/index.md
crutchcorn: I'm gonna remove the mention of `browser engine` since that's a whole other discussion that would go deeper than this post is. It's meant to be a general introduction, not an in-depth technical anaylsis

```suggestion
While moving through the tree, the browser can keep track of the fact that it needs to find an element with the `ID` of `b` and then mark its `<li>` children with a red background. They're "children" because the DOM tree preserves the relationship defined by the HTML.
```

content/blog/understanding-the-dom/index.md
crutchcorn: I wouldn't specifically designate `~` as "more sophisticated.

crutchcorn:
```suggestion
This tree relationship also enables CSS selectors such as the [general sibling selector (`~`)](https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator) or the [adjacent sibling selector (`+`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator) of find siblings to a given selector.
```

content/blog/understanding-the-dom/index.md
crutchcorn: I like the generalizations of your suggestion here, but it removes some of the context I want to leave in:
```suggestion
> The answer behind that? Performance. The [W3 Consortium](https://www.w3.org/Style/CSS/#specs) (the organization that maintains the HTML and CSS standard specifications) points to the tree structure of the DOM and the algorithm used by the browser to traverse the DOM (or, "visit" the nodes in order to figure out what CSS to apply) as not being performant when allowing parent selectors.
```

content/blog/understanding-the-dom/index.md
crutchcorn: Minor edit to include context back:
```suggestion
HTML, as a specification, has tons of tags that are able to be used at one's disposal. These tags contain various pieces of metadata internally to provide information to the browser about how they should be rendered in the DOM. This metadata can then be handled by the browser how it sees fit; it may apply default CSS styling, it may change the default interaction the user has with it, or even what behavior that element has upon clicking on it (in the case of a button in a form).
```

content/blog/understanding-the-dom/index.md
crutchcorn: Awesome catch

PR #125, opened 2019-12-10
Integration merge 12/10/19

crutchcorn:


src/templates/blog-post.js
```diff
@@ -40,6 +44,8 @@ const BlogPostTemplateChild = props => {
 		}, 600);
 	}, [currentTheme, post.frontmatter.title, siteData.siteUrl, slug]);
 
+	console.log(disqusConfig.url);
```
evelynhathaway: `console.log` in prod?

crutchcorn: Removed, great catch

PR #132, opened 2019-12-29
Align post tags to be more inline with other posts

crutchcorn:
For all of our other posts, we have lowercase tags. Because of this post's deviance from that pattern, it was listing some of the values twice. 

Additionally, this removes one of the tags as it didn't seem to match up to an actual skillset someone might be searching for via tag search


crutchcorn:
@evelynhathaway We should, I totally agree. This is just short-term fix

PR #147, opened 2020-02-18
Small CSS Adjustment

MDutro:
The UU logo had a margin-right set at the small break point. I removed that line which seems to have fixed the problem. The logo appears to be behaving itself at other sizes and I did not see any side effects from removing the margin.


crutchcorn:
Do you wanna add yourself as a developer for the site? It's just a way of recognizing your support for the site by adding `"developer"` to the `unicorns.json` file.

PR #218, opened 2020-04-29
Update unicorns.json

thodges314:
I'm creating a profile so that I can add blog content.  I will provide an image shortly.


crutchcorn:
We tend not to update unicorns.json file until we have content associated with the author. Let's make this PR not only updating the unicorns.json file, but adding in the new post in the same PR :)

crutchcorn:
Closing in favor of #219

PR #230, opened 2020-05-18
Integration merge 05/19/20

crutchcorn:


skatcat31:
this is just the integration merge to mainline isn't it?

crutchcorn:
@skatcat31 this is the merge to deploy. We have auto hooks that once a merge is made against `master` it deploys to the site

skatcat31:
so yes. Which makes me curious why it needs a review

crutchcorn:
Prevents rogue contributors and separates team-members' permissions a bit more. It requires one approval from a devops person (which is right now either @evelynhathaway or myself)

skatcat31:
I thought the merge system could be set up so that if it were opened by a specific developer it would auto merge and only those devs could open merges to that branch?

crutchcorn:
They could, but this is meant as an added layer of security. I try my best to make sure that the project isn't just mine and my own work is as scrutinized as any other contributor's. This is just one way I've tried to set this up

PR #247, opened 2020-07-12
File extensions article

skatcat31:
this is gonna need some review. I'm not too happy with that ending, but I'm not exactly big braining right now.


content/blog/what-do-files-extensions-do/index.md
```diff
@@ -0,0 +1,36 @@
...
+}
+---
+
+> A filename extension or file type is an identifier specified as a suffix to the name of  computer file. - [Wikipedia](https://en.wikipedia.org/wiki/Filename_extension)
+
+A long & terse explanation of a file extension exists but to boil it down into simpler terms the file extension is used by a computer to check against a registry of programs to see if a program is registered on the system that can open the file. While there are some differences based on how they are treated by the operating system you use, in most cases, they are used as a simple check to allow a system to see what program can open the file.
+
+# Viewing the File Extension
+
+When most people look at a file they probably don't see a file extension after the name. Often, by default, the file extension is hidden from the user. You can easily find articles that outline how to turn it back on again, however, for longevity, we won't be linking any as with time they can become obsolete. Now when looking at the file its extension will be visible right after the name.
```
crutchcorn, line 20: > When most people look at a file they probably don't see a file extension after the name. Often, by default, the file extension is hidden from the user.

I think this would be a good place to showcase this in a screenshot. I can supply one for Windows/Mac. What do you think?

skatcat31: The problem is that information can become out of date. It's changed for windows 10, it changed for windows XP, it changed for windows 9*, it changed in windows 3.1.1, it was different for NT... it's all over the place where you'll find it

crutchcorn: To counter, that's going to be everything on this blog. Everything will eventually outdate itself. I'd much much much rather add images and references that will become outdated that might even help a single person learn the concept better.

It's not as if the blog is static in it's content. We can always come back and update

skatcat31: This information becoming out of date would be fairly surprising to me. It's easier for an OS to index a filename than it is to index the file name and the magic bits, leading to much faster program launches, but that information could still help people. Sure we can go through with that.

PR #265, opened 2021-03-09
Gatsby 3 upgrade

crutchcorn:
This PR is an initial migration to Gatsby 3, Husky 5, ESLint 7, and more

While both serving and building seem to work as-intended, I noticed that the build includes some problems with images not displaying when they should. Otherwise, this PR is perfectly functioning.

~~It does not include the migration to `gatsby-image-plugin`, which will be in a future PR~~ EDIT: This PR now includes this migration


skatcat31:
haha it's been heiight in source for awhile... aiight

but yeah mostly linting changes. The core stuff looks good to go.

Issue #282, opened 2021-09-23
Tabs


crutchcorn:
I want to be able to have tabs of content that allow you to select multiple different items of context

https://github.com/jhildenbiddle/docsify-tabs

https://github.com/jthegedus/docsify-select

crutchcorn:
https://astexplorer.net/#/gist/d9029a2e8827265fbb9b190083b59d4d/7eef77d0d6caa2a504e6b56d556a89b36d434810

https://github.com/mrzmmr/unist-util-find-all-between

https://www.gatsbyjs.com/tutorial/remark-plugin-tutorial/#what-you-will-learn-in-this-tutorial

crutchcorn:
https://www.gatsbyjs.com/plugins/gatsby-transformer-remark-rehype/

https://www.gatsbyjs.com/plugins/gatsby-remark-component-parent2div/

https://github.com/rstacruz/gatsby-remark-component

https://github.com/rehypejs/rehype-react

crutchcorn:
Remark === Micromark under-the-hood

https://github.com/remarkjs/remark-frontmatter

https://github.com/micromark/micromark#extensions

https://github.com/micromark/micromark#htmlextension

https://github.com/micromark/micromark-extension-frontmatter/blob/main/dev/lib/html.js

https://github.com/micromark/micromark-extension-frontmatter/blob/main/dev/lib/html.js

crutchcorn:
[`remark-frontmatter`](https://github.com/remarkjs/remark-frontmatter) -> [`micromark-extension-frontmatter`](https://github.com/micromark/micromark-extension-frontmatter) -> [`mdast-util-frontmatter#frontmatterFromMarkdown`](https://github.com/syntax-tree/mdast-util-frontmatter) -> Other plugins -> [`mdast-util-frontmatter#frontmatterToMarkdown`](https://github.com/syntax-tree/mdast-util-frontmatter)

crutchcorn:
https://github.com/remarkjs/remark/discussions/869

PR #285, opened 2021-10-22
chore: add functional migration to our internal Gatsby Plugins

crutchcorn:
This PR upgrades internal Gatsby plugins to use UnifiedJS new plugins in preparation for a custom Remark plugin to add tabs: #282

However, part of this upgrade is that [UnifiedJS has gone all-in on ESM](https://github.com/syntax-tree/unist-util-visit/issues/23#issuecomment-825010119)

Unforunately, [Gatsby has themselves not done the same](https://github.com/gatsbyjs/gatsby/discussions/31599) and even, as of current, even Gatsby 4 does not support ESM

This PR upgrades these deps and uses a workaround to include ESM modules in the plugin system anyway.

However, doing so seems to've drastically slowed down our build speed


crutchcorn:
Closing in favor of #289

PR #286, opened 2021-11-04
feat: Add suggested articles sidebar on desktop

crutchcorn:
Sooooo some usage stats from our analytics show that 5x more folks read on our site on desktop as they do mobile. This seems consistent w/ other tech blogs of folks I've talked to

Shows three related articles on the right side of the screen

This provides us a way to suggest other articles for people to read. Hopefully, it'll allow folks to find content that's related to what they're looking for

This suggestion "algorithm" is home-grown and works like so:

```
Get 3 similar articles to suggest in sidebar.
Base off of article series,and similar tags.
If neither apply, simply grab latest articles

However, they should take the precedence. If there are
series articles, they should suggest higher than
matching tags

We check exactly how similar tags are in general. For example, given one
post with 4 tags that match, and another post with only 2, the one with
4 tags will show above the one with 2.

For suggested articles, get the articles only within
1 series order of each other.

So, if we got "2", we could get:
1, 3, 4

But not:
1, 3, 5

Or, alternatively, if we got "3", we could get:
1, 2, 4

But not:
1, 4, 5
```

This may add to our build time by a fair amount, but it's not like it's too bad currently and in order for it to get too much worse as a result of this change, we'd need a lot more articles


crutchcorn:
Don't merge - it seems there's an issue with the first page building wrong. Investigating soon, but not too soon

crutchcorn:
This PR is now built on top of #289

crutchcorn:
Had Twitch chat and multiple others test this fairly well and we have extensive tests. Merging!

PR #288, opened 2021-11-13
Add my first article

Rodentman87:
I think I got this all good to go, I ran it locally and clicked around and it all seems to be working right. I wasn't sure whether or not I should use the original publishing date or today since it's being added to this site now. I put the original date for now so as to keep it in sync with the publish date of the original article, let me know if that should be switched.


crutchcorn (changes requested):
Thanks so much for contributing this piece! :D 

This PR looks good to me. Before we merge, let's go ahead and update the publish date though. We try to keep it consistent with when we "launch" a post on the site - even if cross-posted. This makes it show up in the most recent posts, and adds it to RSS reader subscribers.

Because `originalLink` is set, it shouldn't impact SEO (which is the only other thing `publish` changes), so I think we should be good

If we can set the date to Nov 13th 2021, we'll get this merged right away and trigger a build to add to the site

Rodentman87:
Sure thing, should be all up to date (pun intended) now!

crutchcorn:
Once this job finished in CI, your post will be live!

https://github.com/unicorn-utterances/unicorn-utterances/actions/runs/1456111240

It's currently 12:30PM here - so I'll try to make an announcement post on our Twitter!

Issue #292, opened 2021-12-29
CD Facing 503 "Slow Down" errors occationally


crutchcorn:
Occasionally, our CD builds hit a 503 "slow down" from DO spaces:

https://github.com/unicorn-utterances/unicorn-utterances/runs/4656310090?check_suite_focus=true

The S3 Sync command doesn't mention a timing thing anywhere in their docs:

https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html

https://github.com/jakejarvis/s3-sync-action

evelynhathaway:
DigitalOcean recommends an exponential backoff algorithm in response to rate limiting. AWS CLI for S3 appears to implement a solution that should work with DO Spaces, but maybe we're running into the max retries before we backoff enough for the amount of files we're pushing. I'll continue to look into this.

AWS CLI Docs: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html

evelynhathaway:
This should be properly fixed now from #295. I will re-open if it still occurs.

PR #294, opened 2022-01-05
chore: upgrade Node.js, actions, and increase sync retries for backoff

evelynhathaway:
Closes: #292

### DIFF
--- a/.github/workflows/deploy-beta.yml
+++ b/.github/workflows/deploy-beta.yml
@@ -12,7 +12,7 @@ jobs:
       CI: true
     steps:
       - name: Create GitHub deployment
-        uses: bobheadxi/deployments@v0.4.3
+        uses: bobheadxi/deployments@v0.6.2
         id: deployment
         with:
           step: start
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy-beta.yml
```diff
@@ -12,31 +12,20 @@ jobs:
...
         id: deployment
         with:
           step: start
           token: ${{ github.token }}
           env: beta
 
       - uses: actions/checkout@v2
-      - name: Use Node.js 14.x
-        uses: actions/setup-node@v1
+      - name: Use Node.js 16.x
```
evelynhathaway, line 23: Node.js v16 is now the latest LTS, so let's adopt that as our CI runner

.github/workflows/deploy-beta.yml
```diff
@@ -12,31 +12,20 @@ jobs:
...
-        uses: actions/cache@v2
-        env:
-          cache-name: cache-node-modules
-        with:
-          path: ~/.npm # npm cache files are stored in `~/.npm` on Linux/macOS
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
-          restore-keys: |
-            ${{ runner.os }}-build-${{ env.cache-name }}-
-            ${{ runner.os }}-build-
-            ${{ runner.os }}-
```
evelynhathaway, line 39: The cache step is now an option in actions/node@v2 when using `with: cache: npm`

https://github.com/unicorn-utterances/unicorn-utterances/runs/4717609957#step:7:15

Issue #304, opened 2022-01-09
"Articles" text in a specific author's page is plural when only one article exists


Rodentman87:
Visible here: [https://unicorn-utterances.com/unicorns/maisydino](https://unicorn-utterances.com/unicorns/maisydino)

Fairly minor bug, may also apply to the "words" text but I don't think that's likely to ever hit exactly 1.

bobrossrtx:
I will try and fix this right away 😃

bobrossrtx:
Fixed the issue
![Screenshot 2022-01-16 at 00 57 43](https://user-images.githubusercontent.com/73446766/149643239-0453cc0d-91cc-47b7-b971-f2a01876b00e.JPG)

bobrossrtx:
There is a new pull request for the fix, as the other one was before the rebase.

crutchcorn:
Fixed by @bobrossrtx !

PR #306, opened 2022-01-16
Article - How to get started with .NET && Fixed social text on blog profiles & Non-plurel article count.

bobrossrtx:
I fixed the text padding on the social buttons for the Blog Profiles.
I also fixed the issue #304 - It is no longer plurel when there is only one article.


bobrossrtx:
I added an article to the pull request

crutchcorn:
@bobrossrtx Let's break this into three different pull requests if you're okay with it :) 

1) For the "articles" vs "article" fix
2) For the spacing `0.2rem` fix
3) For the new article

Thanks so much for contributing! Love these! Just want to make sure we're reviewing properly!

PR #307, opened 2022-01-17
Fix: fixed padding & margin on social text on profiles

bobrossrtx:
I fixed the padding of the social buttons on each persons profile


crutchcorn:
I hate to bug you about this @bobrossrtx - but we **just** (like 5 minutes ago) merged a major rewrite of the site into `main`. Can you rebase off of that, please? 

The new file should be here:

https://github.com/unicorn-utterances/unicorn-utterances/tree/main/src/page-components/unicorns

PR #308, opened 2022-01-17
Fix: Profile articles become non plurel if less than 2 articles

bobrossrtx:
Articles on person's profile is no longer plurel if they only have one article


crutchcorn:
I hate to bug you about this @bobrossrtx - but we **just** (like 5 minutes ago) merged a major rewrite of the site into `main`. Can you rebase off of that, please? 

The new file should be here:

https://github.com/unicorn-utterances/unicorn-utterances/tree/main/src/page-components/unicorns

PR #311, opened 2022-01-17
Fix: fixed padding & margin on social text on profiles

bobrossrtx:


src/page-components/unicorns/profile-header/profile-header.module.scss
```diff
@@ -110,7 +110,8 @@
 	}
 
 	.socialText {
-		margin-left: 16px;
+		margin-left: 0.5em;
```
crutchcorn, line 113: Let's change this to `rem` - I don't like/trust `em`

```suggestion
		margin-left: 0.5rem;
```

src/page-components/unicorns/profile-header/profile-header.module.scss
```diff
@@ -110,7 +110,8 @@
...
+		padding: 0.2em 0;
```
crutchcorn, line 114: Instead of doing this, let's set:

```css
.socialBtnLink > a {
    align-items: center;
}
```

Which also solves the problem, but doesn't rely on hardcoding padding values

crutchcorn (changes requested):

crutchcorn:
Thanks for this! Merged!

PR #312, opened 2022-01-17
Fix: Profile articles become non plurel if less than 2 articles

bobrossrtx:


src/page-components/unicorns/word-count/word-count.tsx
```diff
@@ -13,7 +13,7 @@ export const WordCount = ({
   return (
     <div className={wordCountStyles.container}>
       <FeatherIcon className={wordCountStyles.icon} />
-      <p>{numberOfArticles} Articles</p>
+      <p>{numberOfArticles} {numberOfArticles < 2 ? 'Article' : 'Articles'}</p>
```
crutchcorn, line 16: This actually has a similar problem, but with `0` articles, like this page:

https://unicorn-utterances.com/unicorns/adueppen

"0 articles" is correct as opposed to "0 article"

```suggestion
      <p>{numberOfArticles} {numberOfArticles === 1 ? 'Article' : 'Articles'}</p>
```

crutchcorn (changes requested):
Almost got it! We just need a small change here

crutchcorn (approved):
Thanks for making that change!

PR #313, opened 2022-01-17
Sort and cull tags for blog posts

crutchcorn:
@evelynhathaway pointed out to me how chaotic our tags list are on our homepage

image

They're not sorted, there are weird tags that don't have any other content in them, and are generally not that helpful

I think a good way to fix that would be to keep our list as reasonably small as possible and simply monitor the list going forward.

After this was done, the list seems much more manageable

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -107,7 +107,7 @@ Now that we've placed the file in the correct location, we need to add metadata 
     published: '2019-10-07T22:07:09.945Z',
     edited: '2020-02-02T22:07:09.945Z',
     authors: ['edpratti'],
-    tags: ['android', 'design', 'figma'],
+    tags: ['android', 'design'],
     attached: [],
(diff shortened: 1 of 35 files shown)


skatcat31:
I don't see any issues with it. The tag changes make sense and are to simplify the model for filtering.

skatcat31 (approved):

fennifith:
The vercel preview is super neat! I also approve of the tag changes, this looks a lot more organized.
I'm noticing that some of the tag filters in the preview (and in production) aren't showing any results even though there are posts tagged with them, which should probably be a separate issue - aside from that, this looks good to me!

fennifith (approved):

crutchcorn:
Turns out that I shouldn't've changed `cpp` to `c++`. Lunr (our search indexer) doesn't like symbols. I'm changing that back and migrating `c#` to `csharp`

Issue #24, opened 2019-08-11
Switch to an in-house or otherwise more reputable comments system


fennifith:
[This blog post](https://fatfrogmedia.com/delete-disqus-comments-wordpress/) is a start on reasons not to use Disqus, but I'll sum up a few of the vital points:
* it absolutely murders the page's overall load time
* the blocky and (IMO) subpar design doesn't match the rest of the site
* the site (and its users) are inhibited by Disqus's [privacy policy](https://help.disqus.com/en/articles/1717103-disqus-privacy-policy)
  * What this means for end users is that there is no way to ensure that their identifying data or usage information is not collected (short of blocking disqus.com with a browser extension, like I have done - though others might not have the time or knowledge necessary to do so).
  * Disqus's privacy policy states that they can collect and share aggregated statistical or demographic data for any purpose. They can (and probably do) also collect identifying data such as IP addresses and "unique Cookie ID" regardless of whether the user actually interacts with the comments widget or not. This leads me to believe that they are building a profile on every user that touches a site with their comments widget in, for the purpose of targeted advertising and other marketing techniques that [privacy advocates commonly disapprove of](https://privacyinternational.org/topics/adtech).

I believe we could create a faster, more convenient, and frankly just better user experience with our own comments system, or maybe using an existing project that we can self-host - but the end result is that we (who I hope are a trustworthy party) are in control of the user data, and not a company that profits from selling it.

crutchcorn:
I don't inherently disagree. I really don't. However, building our own comment system relies on the ability to have an interactive server, something I'm unsure we're going to be taking on our own. We'll also want to build out our own account system as opposed to using JSON that should be taken into account as well, and when we do so we should likely migrate away from Gatsby to our own handrolled nextjs + GraphQL setup. Lots of groundwork should be done before we consider this

adueppen:
I swear I'm not sponsored by them but this does seem like it could be an interesting thing to try to use Firebase for.

crutchcorn:
I want to avoid Firebase. Once structured on them, it's difficult to migrate away. Plus, it's not an OSS platform

VincentTam:
> I believe we could create a faster, more convenient, and frankly just better user experience with our own comments system, or maybe using an existing project that we can self-host - but the end result is that we (who I hope are a trustworthy party) are in control of the user data, and not a company that profits from selling it.

Hi, this is the maintainer of @staticmanlab.  Static comments like

[...]

PR #314, opened 2022-01-17
fix: adds in privacy error detection and handling

### LINKED ISSUE #287: Add ErrorBoundary to Disqus widget
Some ad blockers and privacy filters block the discus scripts from loading. As such when the user reloads the dark theme, it crashes the site

Let's add an error boundary to the widget to make sure that the page loads properly

crutchcorn:
I use a privacy browser plugin called "[Privacy Badger](https://privacybadger.org/)" from the EFF. It's fairly popular and it's a suggested Firefox plugin.

When changing the theme on UU where Disqus is loaded (like a blog post), it will crash the entire page and leave it blank and unresponsive due to it never actually being loaded and the React component trying to do some cleanup.

This PR adds in an intermediary step and handles the error that occurs during this step, replacing it with a message encouraging the user to take a look at #24

Admittedly, this acts as a bit of fuel to the #24 fire, but this is a short-term solution

Closes #287

### DIFF
--- a/src/components/privacy-error-boundary.tsx
+++ b/src/components/privacy-error-boundary.tsx
@@ -1,0 +1,43 @@
+import { Component, ErrorInfo } from "react";
+
+export class PrivacyErrorBoundary extends Component<{}, { hasError: boolean }> {
+  constructor(props: {}) {
+    super(props);
+    this.state = { hasError: false };
+  }
+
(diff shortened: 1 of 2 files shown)


skatcat31:
pretty simple error boundry. Didn't see any anti-patterns

skatcat31 (approved):

adueppen (approved):

PR #315, opened 2022-01-18
CSS Fundamentals PR

ljtechdotca:
Lemme know if anything is off! Still getting use to this! 🤟


crutchcorn:
@ljtechdotca it seems like you were using an older version of the website. I went and rebased for you.

It also seems like your image was not named the same as the `profileImg` field in `unicorns.json` - so I updated the image name for ya.

I also know that you originally posted your article on your own website via DMs. I added an `originalLink` to canonically link to your site. Should have your site ranked on SEO compared to mine

I also pulled all images to be local to the UU filesystem. We try not to link images externally whenever possible. It's a performance and archival thing, mostly.

Finally, I also updated the `roles` field for you. You originally had it set as `designer, developer`, but these roles aren't to self-describe what you're capable of, but rather what your role is within UU itself.

We would love some help with design and development if you'd like to contribute in those areas, but for now I've set it to `author` since that's what's being added in this PR :)

crutchcorn:
I made a mistake and accidentally committed this directly to `main`.

This PR **was** merged, just not using the `merge` button. Sorry for the confusion

PR #316, opened 2022-01-18
<iframe> CSS improvements

### LINKED ISSUE #299: CoderPad embed doesn't fit in the iframe
:wave: - I noticed on this post https://unicorn-utterances.vercel.app/posts/how-to-upgrade-to-react-18

...that the embedded CoderPad example doesn't fit in the iframe, requiring that I scroll to see it regardless of how big my viewport is (I saw this on Firefox on desktop as well as on mobile in portrait and landscape)

image

fennifith:
This PR (partially) fixes #299 by improving the default appearance of `<iframe>` elements on the site:

- Adds iframe border styling (matching the `<pre>` / code embed styles)
- Sets `width: 100%;` to expand the iframe to fill the post area horizontally, and `min-height: 500px;` vertically
- Refactors all iframes in posts to remove redundant inline styles
- Modifies the MarkdownRenderer to set `width="704" height="500"` (the default iframe size & container width) and `loading="lazy"` on all iframes

This change does not entirely remove the scroll bars from all iframes, but it should help to minimize the issue.

Notes:
- I have scrolled through all the modified posts to check for any wonky iframes, but it'd be good to have a second validation of that
- I've verified the new styles in the current Chromium and Firefox desktop versions. However, mobile browsers and webkit/safari have not (yet) been checked.

### DIFF
--- a/src/global.scss
+++ b/src/global.scss
@@ -277,6 +277,14 @@ $pendIconMarg: #{$baseUnit}px;
       padding-right: 5px;
     }
   }
+
+  iframe {
+    width: 100%;
+    min-height: 500px;
+    border: var(--cardOutlineStyle);
+    border-radius: 8px;
+    overflow: hidden;
+  }
 }
 
 .post-lower-area {
--- a/src/utils/markdown/MarkdownRenderer/media.tsx
+++ b/src/utils/markdown/MarkdownRenderer/media.tsx
@@ -3,6 +3,7 @@ import { useMarkdownRendererProps } from "./types";
 import Image, { ImageProps } from "next/image";
 import Zoom from "react-medium-image-zoom";
 import { getFullRelativePath } from "utils/url-paths";
+import { EMBED_SIZE } from "../constants";
 
 export const getMedia = ({ serverPath }: useMarkdownRendererProps) => {
   return {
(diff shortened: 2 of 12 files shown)


src/utils/markdown/MarkdownRenderer/media.tsx
```diff
@@ -48,5 +49,17 @@ export const getMedia = ({ serverPath }: useMarkdownRendererProps) => {
...
       );
     },
+    iframe: (props: React.IframeHTMLAttributes<HTMLIFrameElement>) => {
+      const { src, ...rest } = props;
+      return (
+        <iframe
+          width={EMBED_SIZE.w}
+          height={EMBED_SIZE.h}
+          loading="lazy"
+          {...rest}
```
crutchcorn, line 59: Smart to add `{...rest}` here. I was just about to request we add ability to overwrite these until I noticed

src/global.scss
```diff
@@ -277,6 +277,14 @@ $pendIconMarg: #{$baseUnit}px;
       padding-right: 5px;
     }
   }
+
+  iframe {
+    width: 100%;
+    min-height: 500px;
```
crutchcorn, line 283: Does it make sense to have `min-height` set here? Shouldn't that be handled by the `height` prop?

crutchcorn (approved):
This looks great overall! Thanks for helping making this so so much cleaner

src/global.scss
fennifith: I believe using `height` would override the `height=` property. My intention was to limit the smallest possible height to `500px`, while still allowing individual iframes to be larger if necessary.

PR #320, opened 2022-01-27
Remove inhouse markdown chain in favor of next-mdx-remote

crutchcorn:


crutchcorn:
I realized that this breaks all of our HTML embeds and comes with some tradeoffs I'm not willing to make. Closing

PR #325, opened 2022-02-09
Add search and filter analytics

crutchcorn:
This PR adds in analytics support for search and filter.

The primary motivation for doing so is that I'd like to see if users are actually utilizing these features, and if so, what for? 

This should help us understand if we need to spend more time on search/filtering and potentially add in server-side support for better search or if this wouldn't be beneficial for our users

Open for discussion about the potential privacy implications and what we should do with this

### DIFF
--- a/src/constants/search-and-filter-context/search-and-filter-context.ts
+++ b/src/constants/search-and-filter-context/search-and-filter-context.ts
@@ -1,14 +1,9 @@
-import {
-  createContext,
-  default as React,
-  useEffect,
-  useMemo,
-  useState,
-} from "react";
+import { createContext, useEffect, useMemo, useState } from "react";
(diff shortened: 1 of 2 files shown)


crutchcorn:
In particular, I'd love to hear @evelynhathaway @adueppen and @fennifith's thoughts on this PR. I know you're all fairly privacy minded (as am I) and would love to hear pros/cons

skatcat31 (approved):

evelynhathaway (approved):
I think the search and tag analytics is very minor privacy-wise and is likely expected since filtering and searching for larger websites is done server side, which is recorded as a new page view or event automatically in most analytics suites for blogs.

The code looks great to me too.

PR #350, opened 2022-05-20
[WIP] Shrink client bundle size

crutchcorn:
# Before 

> 834kb parsed, 250kb gzipped

image

image


crutchcorn:
On the homepage, it seems to drop our TTI from 3.6s to 1s. Huge improvement there despite no jump in lighthouse score

image

On the other hand, it raised our lighthouse score on LJ's CSS post page by 10 points. This was thanks to a TTI 0.7s better and almost half the blocking time

image

PR #351, opened 2022-05-22
Add hidden blog post code snippets to social image backgrounds

fennifith:
This PR replaces the background used by the social image generator with a set of CSS styles intended to replicate its appearance. By doing this, the generator can insert code snippets from each blog post in the background of its social media preview. I don't think anyone will even notice this feature, _and that makes it even better._

Here is a comparison between the current image and the new version:
| Current image (with stock background) | New image (with HTML background) |
|--------------------------------------------|------------------------------------------|
| image | image |

In each post, all code snippets are combined into the same element. In the event that a post does not have any/enough code to fill the window, the source code of the preview function is also appended.


crutchcorn:
This is amazing work!

One of my concerns with the new image preview, however, is that the text is more straightened than the original.

This makes the background text more pronounced than the original image and distracts a bit from the title. Can we make the background rotated more similarly to how the original was done?

PR #368, opened 2022-06-15
Create distinct styles for in-text code snippets

fennifith:
Currently, inline code snippets are a bit hard to separate from the rest of the text in a post. This isn't much of an issue in cases where the syntax is easily recognizable, but it can create confusion when the code snippets start to blend in with a larger paragraph.

For example, the WIP "Minecraft Data Pack Programming" series includes the following sentence:
> If you try to flip the order those two subcommands, `at @a as @s` won't actually select the right entity.

Here's how that looks on the site:
2022-06-15-181046_781x64_scrot

In comparison, this is how it looks with the suggested styling in this PR:
2022-06-15-181313_768x70_scrot

### DIFF
--- a/src/constants/theme.ts
+++ b/src/constants/theme.ts
@@ -25,6 +25,7 @@ export const COLORS = {
     dark: "0px 2px 4px rgba(0, 0, 0, 0.27), inset 0px 1px 0px #435e75",
   },
   codeBlockBackground: { light: "white", dark: "#202746" },
+  codeInlineBackground: { light: "#c6dbe5", dark: "#1d495e" },
   //code styles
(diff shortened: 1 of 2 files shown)


crutchcorn:
This looks awesome. One minor concern: have we checked contrast levels?

fennifith:
Yep - here are the contrast ratios for both light/dark theme:
- Light theme: 14.67 contrast ratio
- Dark theme: 9.69 contrast ratio

The text colors have some transparency in blockquotes, so I also checked the contrast for those:
- Light theme: 5.85 contrast ratio
- Dark theme: 5.04 contrast ratio

PR #370, opened 2022-06-16
Create distinct styles for collapsible <details> elements

fennifith:
I'm not aware of any posts currently using this element, but it could be useful for e.g. providing a problem/challenge with a hidden "solution" that shouldn't be immediately available to the reader.

This PR adds a border to show a clear separation between text in expanded details and the rest of the page. This matches the styling currently used in the tabs component.

2022-06-16-095031_822x333_scrot

The touch target for the `<summary>` element is also expanded by the padding on this element (i.e. it extends to the borders of the page) making it somewhat easier to press.

### DIFF
--- a/src/global.scss
+++ b/src/global.scss
@@ -294,6 +294,24 @@ $pendIconMarg: #{$baseUnit}px;
     border-radius: 8px;
     overflow: hidden;
   }
+
+  details {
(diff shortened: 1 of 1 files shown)


crutchcorn (approved):
Looks awesome, glad you're thinking of it! Thanks!

Issue #356, opened 2022-06-02
Add navigation buttons at the end of an article 


crutchcorn:
Currently, when you reach the end of an article that has a series associated with it, we should provide an easy way to navigate to the previous/next chapter, like so:

![Two buttons at the bottom of a page to indicate "next" and "previous" sections](https://user-images.githubusercontent.com/9100169/171702343-e21ae387-052b-468a-b31c-e19d94657802.png)

https://bfnightly.bracketproductions.com/chapter_1.html

PR #371, opened 2022-06-17
Add navigation buttons at the end of each article layout

fennifith:
This PR adds "next/previous chapter" buttons at the end of each article in a series, fixing #356 

- Renames the `series-toc/` directory to `series/`
- Moves the title regex to a shared `base.ts` import
- Adds a `series-nav` component for rendering the prev/next post in a series

Screenshots of the added component:
2022-06-16-212736_1264x855_scrot

On smaller screens/layouts, these buttons will "wrap around" onto a new line as such:
2022-06-16-212816_531x362_scrot


crutchcorn:
Per our convo on Discord - let's use empty `div`s here to avoid right-justified "previous" or "next" buttons when on first/last post. Other than that, both the functionality and code quality LGTM

PR #373, opened 2022-06-25
Add ability to generate epub for collections

crutchcorn:
This PR introduces a build step to automatically generate an `epub` for each of the collections we have. This includes metadata construction including cover generation and chapter metadata


fennifith:
This looks almost perfect in terms of implementation! However, I'm noticing some inconsistencies with how the generated epub files are rendering code blocks.

In the screenshot below, it appears to include the language applied to the code block in addition to the code it contains - I believe shiki is generating this in a `<div class="language-id">` that is normally hidden through CSS on the page. It might be good to remove that before it is passed to the epub converter.

2022-06-25-135611_667x505_scrot

Also, for lines of code that extend past the width of the page, the epub does not appear to wrap its content to the next line - instead just cutting off the remaining text. This might be implementation-specific depending on the epub reader, or perhaps intended behavior for using code blocks, but it would probably be better to have it wrap these lines in our case.

If it would be possible to render line numbers as well, this should eliminate any confusion about where this line wrapping occurs.

Otherwise, this looks great!

crutchcorn:
image

How does this look @fennifith?

PR #378, opened 2022-06-30
newblog: when to use hashmap overloop

kaleem68:


content/data/unicorns.json
```diff
@@ -325,6 +372,26 @@
     "pronouns": "he",
     "profileImg": "./jahirfiquitiva.jpg",
     "color": "#3867d6",
-    "roles": ["translator"]
+    "roles": [
+      "translator"
+    ]
+  },
+  {
+    "id": "Kaleem",
```
crutchcorn, line 380:
```suggestion
    "id": "kaleem",
```

kaleem68 (commented):
Looks good now, thanks for correcting.

PR #387, opened 2022-07-10
Table of contents styling improvements

fennifith:
- Offsets all heading depths within the table of contents by the minimum level used in a blog post
  e.g. even if a post only uses `<h3>` and `<h4>`, the table of contents will still start at depth 1 instead of depth 3 (as it does now)
- Adds distinct `:hover` and `:focus` styles to the heading links
- Consolidates styling in the module SCSS (moved from global.scss)

### DIFF
--- a/src/components/table-of-contents/table-of-contents.module.scss
+++ b/src/components/table-of-contents/table-of-contents.module.scss
@@ -46,7 +46,8 @@
 
 .tocH1 {
 	padding-top: 0;
-	margin-top: 12px;
+	margin-top: 16px;
+	border-left: none;
 }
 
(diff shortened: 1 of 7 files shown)


fennifith:
One note about this PR: since the bold selection styles have now been applied to every heading in the table of contents, I'm noticing more frequent content shifts when the selection changes. When the font-weight is changed, the size of the text is increased, causing some headings to wrap to a new line.

We could prevent this by changing the table of contents to a uniwidth font such as [Recursive](https://fonts.google.com/specimen/Recursive), which would not change the character width based on font weight.

PR #390, opened 2022-07-12
Small improvements to markdown styling

fennifith:
These changes focus on some existing styling behavior used in posts:

- The light mode background used for inline code has been improved to match the color scheme a bit better:
  | Before | After |
  | ------ | ----- |
  | 2022-07-12-133928_805x212_scrot | 2022-07-12-133920_791x209_scrot |

- Inline code styles are now also applied when used inside of `<li>` elements
- Table borders are no longer clipped by the edge of the containing element:
  2022-07-12-134218_846x323_scrot
- `<hr>` elements have styling that matches the other border styles
  2022-07-12-134501_846x261_scrot


crutchcorn (approved):
LGTM!

PR #405, opened 2022-07-27
[BLOG] Vue Composition API Inspector

william-lohan:


crutchcorn (commented):
First and foremost: Wow! This is an amazing article that goes super in-depth. I love that you're linking to source code; this is going to be a great resource to link to various Vue aficionados.

That said, I have some thoughts on how we can punch it up and make it a bit easier to quickly grok. Most developers do a strange cursory glance at the content before deciding to read it, and there are some tips and tricks to win them over.

-------

Just a general suggestion for writing developer content: Smaller 1-3 sentence paragraphs reign king.

I found while writing full-time at CoderPad that smaller paragraphs lead to better reader retention and engagement specifically within the developer space.

------

Additionally, developers to take on a much more staunch "Show, don't tell" methodology with their consumed content. Whenever it's possible to demonstrate what you're talking about (ASTs, code samples, etc), I've found they tend to soak that stuff up.

----

Ignoring that, this article is actually pretty easy to follow. All of these suggestions are optional to apply, and we can 100% publish without any of these changes if you'd like :)

william-lohan:
Than you so much for the input. I made some revisions.

PR #416, opened 2022-08-15
Angular extend class article

crutchcorn:
This article outlines how to create a base class in Angular and extend it to share lifecycle methods. 

This is generally ill advised, but I wanted to include a resource to explain _how_ to do something that's _technically_ possible while explaining more-in-depth why you shouldn't

@layZeeDK, I'd love to see a review from you if you're up for it <3


content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,757 @@
...
+    <p>The window is {{height}}px high and {{width}}px wide</p>
+  `,
+})
+class AppComponent extends BaseComponent {
+}
+```
+
+This solves the error and now `AppComponent` tracks resizing as-expected!
+
+You'll notice, however, that while `BaseComponent` does have the `implements` keyword, the `AppComponent` does not. While it's seemingly not a _requirement_ to have the `implements` keyword on `AppComponent` in modern versions of Angular, it's still highly suggested.
```
LayZeeDK, line 381: Not sure I agree with this.

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,757 @@
...
+
+For example, if we had a template that looked like:
+
+```html
+<base-component></base-component>
+```
+
+We wouldn't get a compiler error, but would have a loose bit of code running needlessly. Ideally we'd like to have `BaseComponent` still able to use lifecycle methods without registering a new template tag.
+
+Fortunately, that's possible, as of Angular 9; simply remove `BaseComponent`'s `@Component` `selector` property and it won't register a new tag.
```
LayZeeDK, line 408: We could also mark it as an abstract class.

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,757 @@
...
+}
+```
+
+That solves one problem, but still leaves one present with using `@Component`: you must add a declaration of the `BaseComponent` into an `NgModule`. Otherwise, you'll end up with the following error during compilation:
+
+```
+BaseComponent is not declared in any Angular module 
+```
+
+Luckily, [since Angular 10 you can now use `@Injectable` to declare your `BaseComponent` instead](https://angular.io/guide/migration-injectable). This sidesteps the problem because `Injectable`s do not need to be declared:
```
LayZeeDK, line 425: Now wait a minute! This is starting to feel very hacky. Services do not have an OnInit hook! If this works it's because this class is incidentally extended by a component sub-class.

LayZeeDK (changes requested):
Reviewed first half.

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,842 @@
...
+There is a better way to write this code differently today that solve the problems of maintainability a bit better.
+
+Namely, using an `@Injectable` class that's provided on a per-class level enables you to have explicit `setup` and `takedown` functions that you call manually:
+
+```typescript
+@Injectable()
+class WindowSizeService {
+  height = window.innerHeight;
+  width = window.innerWidth;
+
```
william-lohan, line 798: window!: Window;
  constructor(@Inject(DOCUMENT) private document: Document) {
    this.window = document.defaultView!;
  }

william-lohan:
Been a while since I've done angular but looks right (also been down the road).

maybe add

```
  window!: Window;
  constructor(@Inject(DOCUMENT) private document: Document) {
    this.window = document.defaultView!;
  }
```

to the final Service solution to be proper

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,842 @@
...
+class AppComponent extends BaseComponent implements OnInit {
+  override ngOnInit() {
+    console.log("And I am the AppComponent")
+  }
+}
+```
+
+The downside here, however, is that `ngOnInit` on `AppComponent` will no longer call the `BaseComponent`'s `ngOnInit` logic. After all, what if you wanted to simply _add_ behavior to `ngOnInit`, rather than replace it entirely?
+
+Luckily, we can use the `super` keyword to refer to the base class instance and call the original method inside of the overwritten method:
```
LayZeeDK, line 502: praise: Good tip!

LayZeeDK: suggestion: Consider mentioning that if we forget to add `override`, TypeScript will tell us that we're attempting to override an existing method on a base class and thus we must mark it with `override`. This in turn tells us that we should probably add a `super` call.

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,842 @@
...
+})
+abstract class BaseComponent implements OnInit, OnDestroy {
+  window!: Window;
+  constructor(@Inject(DOCUMENT) private document: Document) {
+    this.window = document.defaultView!;
+  }
+}
+```
+
+Because of this, this is the recommended way to get access to both the `document` and `window` instance inside of an Angular component, even if it's a non-SSR app.
```
LayZeeDK, line 542: question: Where have you seen this as a recommended way of accessing the `window` object?

LayZeeDK: issue: My recommendation is to never access the `window` object directly in any Angular-specific class. Instead, I would suggest adding an `InjectionToken` providing a data structure such as:

```typescript
interface ViewportDimensions {
  readonly innerHeight: number
  readonly innerWidth: number;
}
```

We would provide it something like this:
```typescript
const viewportDimensionsToken = new InjectionToken<ViewportDimensions>('viewportDimensionsToken', {
  providedIn: 'platform',
  factory: (): ViewportDimensions => ({
    get innerHeight() {
      return window.innerHeight;
    },
    get innerWidth() {
      return window.innerWidth;
    },
  }),
});
```

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,842 @@
...
+To solve this problem you can use Angular's dependency injection to inject an instance of `document` to `BaseComponent`, and get access to the `window` through `defaultView` that way:
+
+```javascript
+@Component({
+  template: ''
+})
+abstract class BaseComponent implements OnInit, OnDestroy {
+  window!: Window;
+  constructor(@Inject(DOCUMENT) private document: Document) {
+    this.window = document.defaultView!;
```
LayZeeDK, line 537: suggestion: Avoid using the non-null assertion operator.

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,842 @@
...
+  `,
+})
+class AppComponent extends BaseComponent implements OnInit {
+  // This code doesn't work. Read on to learn why
+  constructor(@Inject(DOCUMENT) private document: Document) {
+    super(document);
+    console.log(document.body);
+  }
+}
+```
```
LayZeeDK, line 694: As of Angular 14, we have a way to circumvent this: We use the `inject` function in the base class:

```typescript
import { inject } from '@angular/core';

@Component({
  template: ''
})
abstract class BaseComponent {
  // The `Document` type annotation is optional as it can be inferred by the `inject` function
  private document: Document = inject(DOCUMENT);
  
  window!: Window = this.document.defaultView;

  // No constructor needed as we use property injection instead of constructor injection
  
  // ...
}
```

which removes the need for a sub-class to inject the dependencies of a base class:

```typescript
@Component({
  selector: 'app-root',
  template: `
    <p>Test</p>
  `,
})
class AppComponent extends BaseComponent {
  // This code now works as the base class doesn't have constructor parameters anymore
  constructor() {
    super();
  }
}
```

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,842 @@
...
+
+Something worth mentioning is that this code _still_ doesn't work. You'll see a compiler error with the following code:
+
+```
+TS2415: Class 'AppComponent' incorrectly extends base class 'BaseComponent'.
+   Types have separate declarations of a private property 'document'.
+```
+
+To solve this, we simply need to make our `BaseComponent`'s `constructor` properties `public` instead of `private`:
```
LayZeeDK, line 738: suggestion: ...or stop using *parameter properties*:

```typescript
@Component({
  template: ''
})
abstract class BaseComponent implements OnInit {
  private document: Document;
  
  constructor(@Inject(DOCUMENT) document: Document) {
    this.document = document;
  }
  
  ngOnInit() {
    console.log(document.title);
  }
}
```

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,842 @@
...
+
+Now that we've learned how to extend base classes in Angular to share lifecycle methods, allow me to flip the script:
+
+**You shouldn't use a base class in Angular**.
+
+<video title="A shocked sock puppet monkey" src="./shocked-monkey-gif.mp4"></video>
+
+Why?
+
+Well, it's often sited by Angular experts that using a base class is brittle and difficult to maintain.
```
LayZeeDK, line 777: suggestion: Correct spelling.
```suggestion
Well, it's often cited by Angular experts that using a base class is brittle and difficult to maintain.
```

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,842 @@
...
+
+For example, let's say that you have a base component that doesn't use dependency injection, but then suddenly need to add dependency injection. What do you do?
+
+Well, you'd have to refactor every instance that you extended that class.
```
LayZeeDK, line 781: The `inject` function addresses this issue but there are trade-offs. It will implicitly introduce a new dependency which might:
- Result in a run-time error because of a missing provider
- Break tests because of a missing provider

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,842 @@
...
+
+  ngOnInit() {
+    this.windowSize.addListeners();
+  }
+
+  ngOnDestroy() {
+    this.windowSize.removeListeners();
+  }
+}
+```
```
LayZeeDK, line 832: Calling any kind of hook manually, that is a cardinal sin or at the very least a code smell in Angular apps 😬

You are correct in using a component-level provided service but use the `OnDestroy` hook and the service constructor as the `OnInit`:
```typescript
@Injectable()
class WindowSizeService implements OnDestroy {
  height = this.window.innerHeight;
  width = this.window.innerWidth;
 
  constructor(@Inject(DOCUMENT) private document: Document) {
    this.window = document.defaultView!;
    this.window.addEventListener('resize', this.onResize);
  }
  
  ngOnDestroy() {
    this.window.removeEventListener('resize', this.onResize);
  }

  onResize = () => {
    this.height = this.window.innerHeight;
    this.width = this.window.innerWidth;
  }
}
```

Alternatively, use an [`APP_INITIALIZER`](https://angular.io/api/core/APP_INITIALIZER) or an [`APP_BOOTSTRAP_LISTENER`](https://angular.io/api/core/APP_BOOTSTRAP_LISTENER) instead of the service constructor.

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,842 @@
...
+  }
+
+  addListeners() {
+    window.addEventListener('resize', this.onResize);
+  }
+
+  removeListeners() {
+    window.removeEventListener('resize', this.onResize);
+  }
+}
```
LayZeeDK, line 811: *The Angular way* would be to expose an RxJS observable in a service, not regular properties:

```typescript
interface WindowSize {
  readonly height: number;
  readonly width: number;
}

@Injectable()
class WindowSizeService {
  private destroy$ = new Subject<void>();

  size$: Observable<WindowSize>;
  
  constructor(@Inject(DOCUMENT) document: Document) {
    const window = document.defaultView!;
    this.size$ = = fromEvent(this.window, 'resize').pipe(
      debounceTime(50),
      map(() => ({
        height: this.window.innerHeight,
        width: this.window.innerWidth,
      }),
      takeUntil(this.destroy$),
    );
  }
  
  ngOnDestroy() {
    this.destroy$.next();
    this.destroy$.complete();
  }
}
```

This way, consumers of the service can react to resize events and subscribing to the observable using the `AsyncPipe` or similar will trigger a change detection cycle on the resize events.

Do all of your examples even work in actual Angular apps? 🤔 I'm not sure resize events trigger a change detection cycle unless using a `window:resize` host listener.

LayZeeDK (changes requested):
Second half reviewed.

content/blog/angular-extend-class/index.md
crutchcorn: All of my examples work in actual Angular apps, yeah. I tested.

That said, this _is_ a much better fix. Adding this code sample and attributing it to you.

Thank you for your help

content/blog/angular-extend-class/index.md
```diff
@@ -0,0 +1,1010 @@
...
+
+And that's it! I hope this has been an insightful look into how you can extend component logic.
+
+And this isn't an area of stagnation within Angular - they're introducing new functionality to share component logic using [the upcoming `hostDirectives` API](https://github.com/angular/angular/pull/46868).
+
+Hey, while you're here - do you want to learn more Angular in-depth like this? Maybe you've been working in Angular for some time and want to learn React or Vue, but not start from scratch?
+
+Check out [my free book, "The Framework Field Guide", that teaches React, Angular, and Vue all at the same time.](https://framework.guide).
+
+> A huge huge thank you to [Lars Gyrup Brink Nielsen](https://twitter.com/LayZeeDK) for helping revise this article, and providing some of the methodologies and code samples present throughout, including the wonderful RxJS code sample.
```
crutchcorn, line 1010:
```suggestion
```

Seems a bit odd to attribute a co-author as-if they weren't a co-author, so I'm removing this. @LayZeeDK put just as much effort into this article as I had! ❤️

PR #426, opened 2022-09-05
CSS stacking context article

crutchcorn:


RobinHerzig:
Thanks for this article! The progression into the more complex stacking contexts examples is smooth, and I enjoyed the humor as well. Also, the diagrams on lines 676 and 756 were especially helpful in understanding the concepts.

I noticed a couple mistakes though, do you mind if I point them out?
Line 562 should say "'Opacity' above 'Absolute'", I believe. I tested it here: https://codepen.io/Praelium/pen/qBYGWEK
Then I also noticed that line 670 has a spacing error in \`header `, but that's not as important.

PR #438, opened 2022-11-14
Add font files to astro branch

fennifith:
This PR adds both `ttf` and `woff2`-compressed fonts to the site, along with their CSS font-face definitions.

It also adds a custom `Cache-Control` header for all font files to the vercel config, to prevent re-fetching these once they are first loaded.


src/layouts/document.astro
```diff
@@ -8,11 +8,6 @@ import "../global.scss";
 	<head>
 		<meta name="viewport" content="width=device-width" />
 		<meta charset="utf-8" />
-		<link
-			rel="preload"
-			as="style"
-			href="https://fonts.googleapis.com/css?family=Asap:400,700%7CWork+Sans:400,500,600,700%7CArchivo:400,500,600%7CRoboto+Mono:400&display=swap"
-		/>
```
crutchcorn, line 15: Yeah! Get outta here Googs

PR #442, opened 2022-11-19
Add 404 & mailing list pages to astro migration

fennifith:
- Port 404 page behavior to astro
- Convert "confirm" and "thanks" mailing list pages to astro
- Fix a redundant error box showing up on mailing list forms (failing to match a CSS `:empty` selector)
  * Since there isn't any convertkit JS included, I'm assuming these containers wouldn't do anything regardless
- To be resolved:
  * [x] ~~The 404 page doesn't actually work on Vercel~~ (now fixed with `cleanUrls: true`)

### DIFF
--- a/src/components/mailing-list/mailing-list-full.astro
+++ b/src/components/mailing-list/mailing-list-full.astro
@@ -16,12 +16,6 @@
 		provide good insights to the content we have.
 	</p>
 	<div data-style="clean">
-		<ul
-			class="formkit-alert formkit-alert-error"
-			data-element="errors"
-			data-group="alert"
-		>
-		</ul>
 		<div data-element="fields" data-stacked="false" class="formkit-fields">
 			<div class="formkit-field">
 				<input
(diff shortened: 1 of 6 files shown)


src/components/mailing-list/mailing-list-full.astro
```diff
@@ -16,12 +16,6 @@
 		provide good insights to the content we have.
 	</p>
 	<div data-style="clean">
-		<ul
-			class="formkit-alert formkit-alert-error"
-			data-element="errors"
-			data-group="alert"
-		>
-		</ul>
```
crutchcorn, line 24: These need to stay here. I'll have a follow-up PR to fix the issue. Let's revert this change for this PR

src/pages/404.astro
```diff
@@ -0,0 +1,52 @@
+---
+import Document from "../layouts/document.astro";
+import SEO from "components/seo/seo.astro";
+import sadUnicorn from "assets/sad_unicorn_2048.png";
+import { Image } from "@astrojs/image/components";
```
crutchcorn, line 5: Let's use `Picture` instead of `Image`. It's better performing

crutchcorn (changes requested):

PR #446, opened 2022-12-03
I had made the PT translation of the abot-us

rmotafreitas:
I have translated the "About Us" markdown file, simple as that 👌


PrattiDev (approved):
It's all good! Thank you for the awesome work.

PR #449, opened 2022-12-05
Static tab heights

fennifith:
This PR aims to automatically expand tab containers to the largest available tab content size when possible - to prevent moving around items on the page when a tab is selected. This should prevent most unintended scroll movement and possibly improve performance (by reducing the scope of layout recalculations when tabs are changed).

This is currently implemented using CSS `display: grid;` to overlap tab panels in the same grid cell, and switching from `hidden="true"` to `aria-hidden="true" visibility: hidden;` when hidden (so that hidden tabs still affect the layout).

However, these changes *will not apply* in the following conditions:
- Any tab has an approx. line height of >30 (determined during markdown rendering)
- The viewport is <=822px width (i.e. a mobile layout, where tab content could wrap and further expand the tab height)

The PR also includes some further optimizations of the tabs script:
- Removed some `querySelector` calls in favor of building a `TabEntry` structure with references to each tab
  - these references were already being kept in-scope and iterated through on load, so there shouldn't be any performance implications here
- Moved "keydown" and "click" listeners to reduce scope (and reuse the same listener between elements)
- Remove recursive `checkElementsParents` fn in favor of [`Node.contains()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains)

### DIFF
--- a/src/styles/tabs.scss
+++ b/src/styles/tabs.scss
@@ -3,11 +3,6 @@
 .tabs {
 	-webkit-tap-highlight-color: transparent;
 
-	@include until(822px) {
-		// Override the js-calculated minHeight when the layout can wrap
-		min-height: 0 !important;
-	}
-
 	&__tab-list {
 		padding: 0;
 		margin: 0 1rem -2px;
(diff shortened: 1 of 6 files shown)


src/utils/markdown/scripts/tabs.ts
```diff
@@ -1,145 +1,129 @@
...
 export const enableTabs = () => {
-	const tabLists = document.querySelectorAll('[role="tablist"]');
+	// Array of all TabName->Element mappings in tab sets
+	const tabEntries: TabEntry[] = [];
+
+	// Handle arrow navigation between tabs in the tab list
+	function handleKeydown(this: HTMLElement, e: KeyboardEvent) {
+		if (e.keyCode === 39 || e.keyCode === 37) {
+			const tabs = this.children;
+			let tabfocus = +(this.dataset.tabfocus || 0);
```
crutchcorn, line 19: Can we change `+numOrString` to `Number(numOrString)`?

crutchcorn (changes requested):

src/styles/tabs.scss
```diff
@@ -55,5 +57,33 @@
 		border: 2px solid var(--darkPrimary);
 		border-radius: 0.5rem;
 		padding: 1.5rem;
+
+		&[aria-hidden="true"] {
+			display: none;
+		}
+	}
+
+	@include from(823px) {
```
crutchcorn, line 66: Also, let's document where this number came from

crutchcorn (approved):

PR #461, opened 2023-01-12
What's an UUID article

crutchcorn:


content/blog/what-are-uuids/index.md
```diff
@@ -0,0 +1,228 @@
...
+> - [X.500 DN](https://en.wikipedia.org/wiki/X.500): `6ba7b814-9dad-11d1-80b4-00c04fd430c8`
+
+These UUID versions output a UUID that contains a hash of the namespace and name concatenated together.
+
+You can think of the generation algorithm for both of these UUID versions as the following:
+
+```javascript
+UUID = hash(NAMESPACE + NAME)
+```
+
```
qarnax801, line 163: **Additional explanation:**
The reason a namespace is needed (instead of just hashing the name alone) is that even if the name is unique within a certain set of items (or "namespace" of items) it's likely that two different items from two different namespaces share the same name.
So to avoid duplicate UUIDs across ALL namespaces, we add the namespace UUID and hash the combination.
_And that allows us to generate the same UUID with the same input each time, which means we don't even have to store the value, we can generate them on the go._

content/blog/what-are-uuids/index.md
```diff
@@ -0,0 +1,228 @@
...
+- A pre-defined UUID "namespace"
+- Any string for a "name"
+
+> The UUID specification establishes 4 pre-defined namespaces for common use cases. The pre-defined namespaces are:
+>
+> - [DNS](https://en.wikipedia.org/wiki/Domain_Name_System): `6ba7b810-9dad-11d1-80b4-00c04fd430c8`
+> - [URL](https://en.wikipedia.org/wiki/URL): `6ba7b811-9dad-11d1-80b4-00c04fd430c8`
+> - [OID](https://en.wikipedia.org/wiki/Object_identifier): `6ba7b812-9dad-11d1-80b4-00c04fd430c8`
+> - [X.500 DN](https://en.wikipedia.org/wiki/X.500): `6ba7b814-9dad-11d1-80b4-00c04fd430c8`
+
```
qarnax801, line 155: Could be useful to mention that one can create a namespace UUID if they want, and that these 4 are not the only ones.

qarnax801 (changes requested):
Aside from the few sentence rephrasing suggestions, I also proposed additional explanation on how UUIDv3 and UUIDv5 work

content/blog/what-are-uuids/index.md
```diff
@@ -0,0 +1,228 @@
...
+This means that we can take the previous UUID:
+
+```
+a5abec44-7ce0-437c-972a-cf451b4fde2b
+```
+
+And determine that this is a UUID with the version of `4` and the variant of `9`.
+
+> Speaking of "UUID version numbers", what are those? I know we outlined earlier that there were different ways of generating a UUID; what are they?
```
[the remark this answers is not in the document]
crutchcorn: Strongly disagree. It's a quote from the perspective of the reader.

content/blog/what-are-uuids/index.md
crutchcorn:
```suggestion
>
> You may also use your own hardcoded UUID as a namespace
```

content/blog/what-are-uuids/index.md
```diff
@@ -0,0 +1,228 @@
...
+Likewise, **you should use UUIDv5 when**:
+
+- It's critical that `name` or `namespace` contents cannot be decoded
+- Performance is less of a concern
+
+As UUIDv5's SHA-1 is significantly more secure with hashing its contents.
+
+### Pros of UUIDv3 and UUIDv5
+
+- **Contents are stable**: In instances where you need the input to match the output every time, there are no other UUIDs that can fulfill this use case.
```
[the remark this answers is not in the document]
crutchcorn:
```suggestion
- **Contents are stable**: In instances where you need the output to match the input every time, no other UUID versions can fulfill this use case.
```

PR #472, opened 2023-02-04
Fix iframe embed placeholder size

fennifith:
This PR makes the iframe container sizing consistent with its replacement element, so that no page jumps occur when switching between them.

### DIFF
--- a/src/utils/markdown/rehype-unicorn-iframe-click-to-run.ts
+++ b/src/utils/markdown/rehype-unicorn-iframe-click-to-run.ts
@@ -137,11 +137,9 @@ export const rehypeUnicornIFrameClickToRun: Plugin<
 						"data-pageicon": iframePicture
 							? JSON.stringify(iframePicture)
 							: undefined,
-						"data-width": width,
-						"data-height": height,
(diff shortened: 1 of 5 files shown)


src/utils/markdown/rehype-unicorn-iframe-click-to-run.ts
```diff
@@ -139,7 +139,9 @@ export const rehypeUnicornIFrameClickToRun: Plugin<
 							: undefined,
 						"data-width": width,
 						"data-height": height,
-						style: `height: ${height}px; width: ${width}px;`,
+						style: `height: ${+height ? `${height}px` : height}; width: ${
+							+width ? `${width}px` : width
+						};`,
```
crutchcorn, line 144: Let's use `Number()` instead of `+`:

```suggestion
					style: `height: ${Number(height) ? `${height}px` : height}; width: ${
							Number(width) ? `${width}px` : width
						};`,
```

crutchcorn (changes requested):
Minor thing, but let's change this for readability

crutchcorn (approved):

Issue #474, opened 2023-02-06
Possible typo in "UUIDv3 vs. UUIDv5" section


gitname:
Link to section: https://unicorn-utterances.com/posts/what-are-uuids#When-should-you-use-UUIDv3-vs-UUIDv5

I felt confused reading the following comparison (I highlighted a few words in yellow here):

![image](https://user-images.githubusercontent.com/7143133/217080613-e9186a46-f159-4e3c-a7fb-985507004f93.png)

I am wondering whether, for UUIDv3, the author meant to say that it is not critical that contents _cannot_ be decoded (meaning, the user is OK with contents being decoded).

crutchcorn:
Hi! Thanks for raising this :D I'm the author of this article.

We should totally fix this language. The gist of what I was trying to say is that:

UUIDv3 should be used when it doesn't matter too much if the data can be decoded.
UUIDv5 should be used when it's critical that data cannot be decoded.

Would you possibly be willing to make a PR to fix this language? :D

gitname:
Your clarification makes sense to me. I opened the PR linked above.

PR #475, opened 2023-02-07
Clarify UUIDv3 vs UUIDv5 bullet points (#474)

gitname:


crutchcorn:
Thank you so much! This is a great way of clarifying this

PR #480, opened 2023-02-08
Fix tab margins behavior on FFG page

fennifith:
Fixes a styling issue on the FFG page that caused the static tab heights implementation (#449) to break.


crutchcorn (changes requested):
This removes the "focus on center of tabs" when the tabs list is not small.

This functionality is important because otherwise, like in my book, the changing of tabs drastically impacts the scroll height and such

fennifith:
Leaving a note for future reference - this is waiting on testing for iOS/Safari to ensure that there are no scroll jumps when tabs are selected. Latest Chrome and Firefox versions have not been able to reproduce any scroll jumps.

Link for testing: https://unicorn-utterances-git-fix-tab-margins-on-ffg-oceanbit.vercel.app/posts/angular-extend-class

crutchcorn (approved):

fennifith:
Agree that users could see the extra margin as a bug - I think this would appear more intentional if the margin was inside the tab container, but IIRC we decided against that for other reasons.

Merging for now, and we can revisit the design aspects later if needed.

PR #505, opened 2023-03-07
DDD Article: Correct typo ("bass" -> "boss")

Jutanium:


fennifith:
Ooh, thanks for the fix! :)

PR #506, opened 2023-03-11
Prevent update log spam by replacing rollup-plugin-copy

### LINKED ISSUE #482: Disable [astro] UPDATE text in HMR
https://github.com/withastro/astro/blob/8b7cb64dadfca93c65d62df54754633d398cb2ed/packages/astro/src/core/messages.ts#L49-L51

https://github.com/withastro/astro/blob/8b7cb64dadfca93c65d62df54754633d398cb2ed/packages/astro/src/vite-plugin-astro/hmr.ts#L127

fennifith:
Fixes #482 

Adds [`symlink-dir`](https://www.npmjs.com/package/symlink-dir) as a dependency to prevent update log spam during dev builds.

This should be tested on multiple platforms (MacOS/Windows) to ensure interoperability before merging. (`symlink-dir` does appear to have cross-platform support, but I'm not sure if it'll work with Astro/Vite on all platforms, or if HMR updates are detected)

Tested platforms:
- [x] Linux
- [x] Windows
- [x] MacOS


fennifith:
It seems like Vercel maybe doesn't particularly like this approach, as some of the blog post images are turning into this in their expanded form:
2023-03-12-112013_1235x756_scrot

PR #549, opened 2023-06-09
[UWU] Collection Cards to match Spec

crutchcorn:
This PR updates the collection cards to match spec a bit better.

# Before + After

image


src/components/collection-card/collection-card.astro
```diff
@@ -26,6 +27,8 @@ type CollectionCardProps = {
 
 const { collection, unicornProfilePicMap } = Astro.props as CollectionCardProps;
 
+console.log({ unicornProfilePicMap });
```
fennifith, line 30:
```suggestion
```
Left over debug log, perhaps? ;)

fennifith (approved):

PR #557, opened 2023-06-24
[UWU] Tabs Styling Updates

crutchcorn:
This PR updates tabs to look like this:

image

Per the mockups.

It temporarily removes the `small-tabs` logic, as there's no fixed content width to dynamically calculate from.

Preview: https://unicorn-utterances-git-uwu-tabs-refresh-oceanbit.vercel.app/posts/angular-extend-class

### DIFF
--- a/src/styles/tabs.scss
+++ b/src/styles/tabs.scss
@@ -2,10 +2,8 @@
 
 :root {
 	--tabbed-wrapper_padding: var(--spc-2x);
-	--tabbed-wrapper_corner-radius_inner: var(--corner-radius_l);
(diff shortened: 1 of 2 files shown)


crutchcorn:
Feedback from @PrattiDev 

- [ ] Unselected tab's text should not be selectable, as they are "buttons".
- [ ]  Selected tabs should not hover or pressed states
- [ ]  Corner tokens _inner and _outer have been changed in Figma. Previous tokens were going to lead to failure if the text resized the tabs.
- [x]  ~~The code blocks themselves aren't themed at all, but I suppose that's for another time?~~

fennifith (approved):

PR #580, opened 2023-06-26
[UwU] Misc. spacing, scaling, & centering fixes

fennifith:
- Remove `1fr 1fr` row template, which was forcing both the metadata & bio rows to be an equal height
- Add `align-self: center` to items on the metadata row for vertical alignment
- Fixed scaling bug with a font style incorrectly applied to `:root`
- Replace spacing styles with `var(--site-spacing)` to match specs

### DIFF
--- a/src/styles/text-styles.scss
+++ b/src/styles/text-styles.scss
@@ -33,7 +33,6 @@
 
 :root {
 	font-family: var(--uu-font-family);
-	font-size: 1rem;
 }
 
 h1, .text-style-headline-1 {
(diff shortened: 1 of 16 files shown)


src/styles/text-styles.scss
```diff
@@ -31,6 +31,11 @@
 
 // TODO: Add back fallback font for Figtree
 
+:root {
+	font-family: var(--uu-font-family);
+	font-size: 1rem;
```
crutchcorn, line 36: We need to remove this root font size, as it breaks font scaling otherwise

crutchcorn (changes requested):
Minor change

crutchcorn (approved):

PR #581, opened 2023-06-27
[UwU] Update Post Cards to match Spec

crutchcorn:
This PR:

- Updates the card's styling to match the spec
- Updates all usages of the dense icon to include the intended padding
- Fixes the focus state of chips using `outline` rather than a border

image

### DIFF
--- a/src/components/post-card/post-card.module.scss
+++ b/src/components/post-card/post-card.module.scss
@@ -54,6 +54,7 @@
 	border-radius: var(--card_corner-radius);
 	border: var(--card_border-width) solid var(--card_border-color);
 	background: var(--card_background-color);
+	overflow: hidden;
 }
 
 .postBase:hover {
--- a/src/components/post-card/post-card.module.scss
+++ b/src/components/post-card/post-card.module.scss
@@ -171,7 +171,13 @@
 	margin-top: var(--card_tags_padding-top);
 }
 
-.extendedPostContainer {
+.spacer {
+	height: 1px;
+	flex-grow: 1;
+}
+
+.extendedPostContainer,
+.regularPostContainer {
 	display: flex;
 	flex-direction: column;
 	flex-wrap: nowrap;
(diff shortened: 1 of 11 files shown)


src/components/post-card/post-card.module.scss
```diff
@@ -1,112 +1,229 @@
...
-	&:hover {
-		background-color: var(--surface_primary_emphasis-low);
-		border-color: transparent;
-	}
+.authorList {
+	list-style: none;
+	margin: 0;
+	padding: 0;
+	display: flex;
+	flex-direction: row;
```
fennifith, line 110:
```suggestion
	flex-direction: row;
	flex-wrap: wrap;
	gap: var(--card_detail-row_gap);
```

I believe this is needed to properly handle multiple authors in the list (currently they don't wrap, and have no spacing between authors).

src/components/post-card/post-card.module.scss
```diff
@@ -1,112 +1,229 @@
...
-	margin-bottom: var(--spc-4x);
+.extendedPostContainer,
+.regularPostContainer {
+	display: flex;
+	flex-direction: column;
+	flex-wrap: nowrap;
+}
+
+.extendedPostImageContainer {
+	height: 1px;
```
fennifith, line 187: This is still rendering the image on the mobile breakpoint with a height of `1px` (which is more visible in dark mode...) - I think this should have a `@include from($tabletLarge)` to make it `display: none;` on mobile.

src/components/post-card/post-card.module.scss
```diff
@@ -1,112 +1,229 @@
...
+	height: 100%;
+	width: 100%;
+	position: absolute;
+	top: 0;
+	left: 0;
+	object-fit: cover;
+}
+
+.extendedPostImageBackground {
+	background-color: var(--primary_variant);
```
fennifith, line 226: I think this background can be placed on `.extendedPostImageContainer` for the same result to save a div :)

(I did test this in-browser first)

PrattiDev (approved):
Solid work.

fennifith (approved):

PR #598, opened 2023-07-10
[UwU] Post card ctrl click

### LINKED ISSUE #573: Ctrl+Click on a post card does not have consistent behavior
Currently, the `<a>` title within a post card can be ctrl+clicked to open it in a new tab. However, this also activates the `onclick` and navigates the current page to the clicked post.

Ctrl+clicking elsewhere on the post (outside of the `<a>` tag) only activates the `onclick`, which navigates the current page and does not open a new tab.

The `<a>` element _can_ be right clicked to use "open in new tab" from the context menu, which works as expected. However, right clicking on the post card outside of the `<a>` does not provide these options.

Open to discussion on how these cases should be handled, since the `onclick` event doesn't normally define "open in new tab" behavior to begin with. However, the `<a>` used for post titles *can* still open links in a new tab, so it might be good to prevent the `onclick` from running if the click originated within the `<a>` element.

crutchcorn:
Fixes #573

### DIFF
--- a/src/components/post-card/post-card.tsx
+++ b/src/components/post-card/post-card.tsx
@@ -73,8 +73,7 @@ export const PostCardExpanded = ({
 }: PostCardProps) => {
 	return (
 		<li
-			// @ts-ignore
-			onclick={`location.href='/posts/${post.slug}'`}
+			data-navigation-path={`/posts/${post.slug}`}
 			className={`${className} ${style.postBase} ${style.extendedPostContainer}`}
 		>
 			<div className={style.extendedPostImageContainer}>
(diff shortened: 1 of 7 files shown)


fennifith (commented):
Small issue: ctrl+click or clicking on a link inside of a `data-navigation-path` container now activates the container's onclick, and does not open the nested link

(can be reproduced by interacting with the author link within a post card)

crutchcorn:
@fennifith this issue is now resolved.

In addition, this handler supports:

- Middle-clicking the mouse
    - Preventing scroll on middle-mouse
- Not attaching to `a` or `button` tags
- Supporting nested `data-navigation-path` usage
- Adding an explicit `data-dont-bind-navigate-click` handler

fennifith (approved):

PR #601, opened 2023-07-10
[UwU] Initial search page

crutchcorn:
Preview: https://unicorn-utterances-git-uwu-search-page-oceanbit.vercel.app/search

This PR includes:

- [x] URL-driven search
- [x] Client-side pagination
- [x] Updates to search to include collections
- [x] Mobile and desktop styling
- [x] SEO

Does not include (to follow up in a new PR):
- [ ] Integration tests
- [ ] [A11Y](https://www.scottohara.me/blog/2022/02/05/dynamic-results.html)
- [ ] Initial state
- [ ] Failure state
- [ ] Loading state
- [ ] No results found


src/views/search/search-page.tsx
```diff
@@ -0,0 +1,99 @@
+import { useCallback, useMemo } from "preact/hooks";
+import { Pagination } from "components/pagination/pagination";
+import { PostInfo } from "types/PostInfo";
+import { PostCard } from "components/post-card/post-card";
+import unicornProfilePicMap from "../../../public/unicorn-profile-pic-map";
```
fennifith, line 5: This file doesn't exist on dev builds, so it would be preferred to use `utils/get-unicorn-profile-pic-map` in an astro file and pass this in as a prop

PrattiDev:
# To-do:

- [ ] Header: Fix title and count text styles to `button-regular` instead of `button-large`.
image

- [ ] Header: Unfinished token implementation
`--spc-2x` as opposed to the `padding-vertical` tokens specified.

- [ ] Header: Fix lack of border when scrolling
image

- [ ] Sidebar: Max width value for `desktopLarge` is not being applied; defaulting to `tabletLarge`. 
image

PR #608, opened 2023-07-12
[UwU] Navbar fixes

fennifith:
- Fixes the navbar max-width on the collections & post pages to match mocks
- Adds `size={"l" | "xl"}` to the `<Document>` and `<Header>` components to configure which pages use the full-width navbar
- Changes the search form JS submit listener to HTML form behavior
- Renames the table sticky header observer to "sticky-observer-script" and adds to all pages (used to apply the header's `border-bottom` style when pinned)
- Adds the "dense" search bar input variant (specs still need verified)


crutchcorn (approved):
Amazing work!

PR #658, opened 2023-08-22
[UwU] Add search title text-shadow, fix search filter alignment

fennifith:
- Removes var/token indentation to fix devtools usage
- Adds `flex-shrink: 0;` to filter items to prevent misalignment with wrapping text
- Set `z-index: -1` on hero stickers and add a (hardcoded for opacity) `text-shadow` to the hero title

### DIFF
--- a/src/components/checkbox-box/checkbox-box.module.scss
+++ b/src/components/checkbox-box/checkbox-box.module.scss
@@ -40,6 +40,7 @@
 	display: inline-flex;
 	align-items: center;
 	justify-content: center;
+	flex-shrink: 0;
 }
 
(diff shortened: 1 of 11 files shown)


crutchcorn:
Can we revert [da2a39a](https://github.com/unicorn-utterances/unicorn-utterances/pull/658/commits/da2a39a7dadea6ea271a0a94106dcc20b66eb3ee)? I'd rather us just move the sticker rather than add a blur

crutchcorn (changes requested):
Marking for changes per the note mentioned

crutchcorn (approved):

PR #660, opened 2023-08-22
[UwU] Fix mailing list input layout shift

fennifith:
The modular CSS appears to have a higher specificity in prod than in dev mode, resulting in some unreliable behavior. This fix hopefully prevents any more text-styles related issues involving this.

- All of the text-styles classes now use a `:where()` selector to reduce their specificity
- Any default text styling for elements is removed.
  * i.e. no more `p, .text-style-body-large { ...`. Any `<p>` or `<h#>` element will need to specify which text style class to use.
  * This prevents a `<p class="text-style-body-small">` from being affected by the `text-style-body-large` styling (since the two selectors would conflict if placed inside a `:where()` selector)
  * Selectors in `.post-body` now `@extend` the respective text styles to apply them to post contents
- The `margin: 0;` causing input layout shift on focus has been removed.
  * This was entirely unrelated to the specificity problems, but the `margin: 0;` property in `text-style-body-medium` meant that the layout shift would not occur in dev mode.

### DIFF
--- a/content/site/about-us.es.md
+++ b/content/site/about-us.es.md
@@ -1,4 +1,4 @@
-<p class="text-style-headline-2">Nuestro objetivo es proveer los mejores recursos posibles para aprender cualquier tema relacionado con las ciencias de la computación</p>
+<p><span class="text-style-headline-2">Nuestro objetivo es proveer los mejores recursos posibles para aprender cualquier tema relacionado con las ciencias de la computación</span></p>
 
 Ya sea desde cómo se almacena la memoria en assembly o cómo hacer animaciones complejas en CSS, hasta lo que es un bucle For, queremos brindar recursos que sean acogedores, de apoyo e informativos.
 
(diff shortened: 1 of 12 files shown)


content/site/about-us.es.md
```diff
@@ -1,4 +1,4 @@
-<p class="text-style-headline-2">Nuestro objetivo es proveer los mejores recursos posibles para aprender cualquier tema relacionado con las ciencias de la computación</p>
+<p><span class="text-style-headline-2">Nuestro objetivo es proveer los mejores recursos posibles para aprender cualquier tema relacionado con las ciencias de la computación</span></p>
```
crutchcorn, line 1: I don't understand this change, could you explain more?

crutchcorn (changes requested):
We have introduced a regression - this now layout shifts by 1px

image

content/site/about-us.es.md
fennifith: Since it's inside a `.post-body`, a `p.text-style-headline-2` would have both the `.post-body p` and `.text-style-headline-2` styles. Because of the `:where(...)` change, the `.post-body p` takes precedence (we want `.text-style-headline-2`).

crutchcorn (approved):

Issue #655, opened 2023-08-22
[UwU] (Search page) Mobile filter modal's vertical breakpoint is different from its min-width


fennifith:
There is a small range of viewport sizes where the filter columns no longer shrink horizontally, causing the modal to clip past the edge of its container, before it hits the breakpoint to change to a vertical layout.

https://github.com/unicorn-utterances/unicorn-utterances/assets/13000407/9387bcee-4c01-4238-b871-5373139cbb0c

Issue #645, opened 2023-08-18
[UwU] Collapsible headers need focus/hover/active states in search


crutchcorn:
<img width="170" alt="image" src="https://github.com/unicorn-utterances/unicorn-utterances/assets/9100169/413ed38d-6dfa-454f-a5ba-6ce690b174e3">

Issue #643, opened 2023-08-18
[UwU] Missing focus states on checkbox labels


crutchcorn:
There is hover state, but no focus state:

<img width="186" alt="image" src="https://github.com/unicorn-utterances/unicorn-utterances/assets/9100169/8b490f96-d535-4bff-8d42-b8efe95f106d">

PR #675, opened 2023-09-09
[UwU] Search page styling changes

fennifith:
- Sets the `max-width` of the search page layout to match the "xl" page size.
- Adds `opacity: 0;` to make the sidebar invisible when hidden
- Changes the layout to `display: grid` in order to perform the sidebar transition when hidden
  * This fixes the issue where the page would briefly show the sidebar and then perform the hidden transition when the page first loads
- Adds hover/focus styling to filter elements:
  * #645 
  * #643
- Adds styling to override the input autofill color/background
- Fixes breakpoint comparisons to use separate ranges; so that usage does not overlap at breakpoint values
  * `until(x)` is now `[0, x]` (inclusive) or `width <= x`
  * `from(x)` is now `[x+1, inf]` or `width > x`
- Increases the filter dialog breakpoint by 100px to prevent overflow
  * #655

### DIFF
--- a/src/views/search/components/filter-section.module.scss
+++ b/src/views/search/components/filter-section.module.scss
@@ -38,26 +38,43 @@
 }
 
 .sectionTitle {
+	// Stop being a button 😡
+	all: unset;
+
 	grid-column: 1 / span 2;
 	grid-row: 1;
 
 	display: flex;
 	align-items: center;
 	gap: var(--search-page_filter_list_header_gap);
-	width: 100%;
+
 	// Replace with tokens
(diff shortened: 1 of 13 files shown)


src/views/search/components/filter-section.module.scss
```diff
@@ -38,26 +38,43 @@
 }
 
 .sectionTitle {
+	// Stop being a button 😡
```
crutchcorn, line 41: No u

crutchcorn (approved):

Issue #676, opened 2023-09-09
[UwU] Styling for <code> and <kbd> elements


PrattiDev:
![image](https://github.com/unicorn-utterances/unicorn-utterances/assets/17130024/366ce82c-18a2-4810-a765-82af4354901e)

- [x] Styling for `<code>` elements
```
    font-family: var(--uu-font-family-code);
    font-size: 0.8em;
    font-weight: 500;
    color: var(--secondary_on-variant);
    background-color: var(--surface_secondary_emphasis-none);
    padding-bottom: 0.15em;
    padding-left: 0.15em;
    padding-right: 0.15em;
    border: 0.075em solid var(--secondary_default);
    border-radius: 0.25em;
```

- [x] Styling for `<kbd>` elements
```
    font-family: var(--uu-font-family-code);
    font-size: .8em;
    font-weight: 500;
    color: var(--black);
    background-color: #eaeaea;
    text-shadow: 0px 0.075em var(--white);
    line-height: 1;
    padding: 0.15em;
    border: 0.075em solid var(--white);
    border-radius: 0.25em;
    filter: var(--shadow_sticker_drop_shadow);
```

PR #677, opened 2023-09-11
[UwU] Inline <code> / <kbd> styling

fennifith:
Adds the inline markdown styling for `<code>` and `<kbd>` elements from #676

### DIFF
--- a/src/styles/markdown/hints.scss
+++ b/src/styles/markdown/hints.scss
@@ -86,12 +86,10 @@
 
 			& > *:first-child {
 				margin-top: 0;
-				padding-top: 0;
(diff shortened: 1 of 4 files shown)


fennifith:
TODO note for myself:

- https://unicorn-utterances-git-uwu-inline-styling-oceanbit.vercel.app/posts/setup-a-react-native-monorepo
  The "file list" component is using `<code>` elements for all of its filenames, so it's inherited the styling changes in this PR. That should probably be changed to use `<span class="text-style-...">` instead

crutchcorn (approved):

PR #700, opened 2023-09-18
[UwU] List changes

fennifith:
- Removes the bowtie styling from unordered lists
- Inherits the `<p>` font styling for list item content by default
  (this fixes empty `<li>text content</li>` having a different font size from `<li><p>text content</p></li>`)
- Removes the styling preventing nested lists from adding top/bottom margins
- Applies styling to `<ol>` ordered lists as well as unordered lists

### DIFF
--- a/content/blog/test-edge-cases/index.md
+++ b/content/blog/test-edge-cases/index.md
@@ -18,6 +18,12 @@ This is regular text.
 - List item 2
 - List item 3
 
+# Single-level ordered list
+
+1. List without a nested `<p>` tag
+2. List item 2
+3. List item 3
+
 # Single-level padded list
 
 - List item 1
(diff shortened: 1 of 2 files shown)


src/styles/markdown/lists.scss
```diff
@@ -14,32 +14,18 @@
 	);
```
crutchcorn, line 14: Let's remove the `--list_bowtie` variables too? 😄

content/blog/test-edge-cases/index.md
```diff
@@ -0,0 +1,155 @@
+---
+{
+	title: "Test Content: Edge Cases",
+	description: "Look at all those edge cases!",
+	published: '2020-09-18',
```
crutchcorn, line 5: Let's set this to `1999` or something so that we can easily detect if it makes it into prod

crutchcorn (approved):

PR #711, opened 2023-09-21
[UwU] About page redesign

fennifith:
- Combines about page into a single MDX file per translation
- Adds a license component for sticker attribution

---

- **TODO:** code blocks in the license dialog aren't being processed by `unified` / `remark-shiki-twoslash` correctly in the deployment build, resulting in incorrect styling (the "shiki" classes/spans aren't being added). This works perfectly fine in dev mode. I have no clue what's going on.


PrattiDev:
Figma contains updated specs.
• Max-width for the hero banner, the content, and the top header. (All different)
• Column count for all breakpoints
• Grid spacing tokens
• Hero banner functionality

src/views/about/about.astro
```diff
@@ -1,71 +1,64 @@
...
-				))
+				Object.values(data.tags).map((tag) =>
+					"image" in tag ? (
+						<License
+							name={tag.displayName}
+							image={tag.image}
+							url={
+								`https://github.com/${siteMetadata.repoPath}/blob/main/public` +
+								tag.image.replace(".svg", "-LICENSE.txt")
+							}
```
crutchcorn, line 53: Instead of showing all `tag`s, I think we should just show the licenses for the stickers, alongside the usage of `explainer`:

https://github.com/unicorn-utterances/unicorn-utterances/blob/uwu/src/utils/get-stickers.ts#L16-L17

crutchcorn:
# A11Y Audit

- [ ] We need an alt for the header image

image

- [ ] We need to `aria-hide` this icon:

image

- [ ] Add list to translations list with an `aria-labelled-by` the previous text:

image

- [ ] UU socials should be in a list with a label:

image

- [ ] Sponsors should be in a list with a labelled-by the header prior

image

- [ ] Hide image on `info` box:

image

- [ ] Contributors should be in a list with `aria-labelled-by` previous header:

image


- [ ] "View profile" aria-label should have name of contributor to make tab order sound proper:

image

- [ ] Licences should be in a list with a label from the header above:

image

- [ ] "View license" should have name of the item in it to read better in tab order

image

- [ ] `alt=""` the images for the licenses

image

- [ ] `alt=""` the profile picture of contributor, as it's already explained by the name next to it

crutchcorn (approved):

PR #715, opened 2023-09-23
[UwU] Add noindex frontmatter property for test posts

### LINKED ISSUE #699: Post for testing markdown content
Add a "test post" for checking markdown behavior with styling functionality on the site.

This should display all the weird edge cases of using content. Can tabs be placed inside a list? Can inline code be used in a table heading? Ideally, almost every possible weird markdown combination should be here in one place.

This could be useful for both testing behavior and as a reference for the site's markdown components.

The post should not be included in search indexing, in the sitemap, or in post lists anywhere on the site. This could be done by adding a `noindex: true` flag to the frontmatter data.

fennifith:
Fixes #699 - Adds an "example post" that should be excluded from the rest of the site.

- Any `utils/api` queries (and `getAllExtendedPosts`) exclude posts with the `noindex: true` frontmatter.
  * This should result in them being excluded from all post lists and the generated search index
- `noindex` posts are filtered from the RSS feed
- A `<meta name="robots" content="noindex, nofollow" />` is added to the header of any noindex post page.
  According to Google's [developer docs](https://developers.google.com/search/docs/crawling-indexing/block-indexing), the page should still be accessible from `robots.txt`; i.e. adding this tag alone should be sufficient.

TODO: not sure if anything needs to be done to the sitemap

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -45,6 +45,15 @@ export default defineConfig({
 					return prev;
 				}, {}),
 			},
+			filter(page) {
+				// return true, unless lart part of the URL ends with "_noindex"
+				// in which case it should not be in the sitemap
+				return !page
+					.split("/")
+					.filter((part) => !!part.length)
+					.at(-1)
+					.endsWith("_noindex");
+			},
 			serialize({ url, ...rest }) {
 				return {
 					// remove trailing slash from sitemap URLs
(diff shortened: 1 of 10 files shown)


crutchcorn:
Sitemap changes should not be necessary. Google will know about this page, but won't index because if the meta tag

crutchcorn (approved):

PR #722, opened 2023-09-25
[UwU] Locale fixes

### LINKED ISSUE #716: Index localized pages as adjacent links in the HTML header / sitemap
According to google's docs, there are a few things the site should be doing to index its translated pages: https://developers.google.com/search/docs/specialty/international/localized-versions#all-method-guidelines

- While we provide HTML `<link rel="alternate">` header tags for translations, it excludes the link entry for the current page, whereas the docs specify:
  > one link for each page variant **including itself**.

  > If two pages don't both point to each other, the tags will be ignored. This is so that someone on another site can't arbitrarily create a tag naming itself as an alternative version of one of your pages.

  I think this means that localized pages are not currently being indexed correctly. (this can be tested on the [hreflang tool](https://technicalseo.com/tools/hreflang/) linked in their docs)

- Additionally, the language attribute is declared as `href-lang=""`, whereas it should be [`hreflang=""`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#hreflang) (with no hyphen - I'm not sure if page crawlers can still resolve the attribute, but it should probably be fixed regardless)

- **TODO:** it is unclear if `hreflang="x-default"` should be provided alongside an additional `hreflang="en"` entry, since those would be the same link in our case.

  Looking at other sites:
    - MDN does not specify an `hreflang="x-default"` and only provides an `hreflang="en"`
    - Google's docs site provides both `hreflang="x-default"` and `hreflang="en"`, which have the same URL

  So it seems like UU could do either.
  
- Additional `<xhtml:link rel="alternate">` links can be added to the sitemap for these pages.

  This may be as simple as providing the [i18n](https://docs.astro.build/en/guides/integrations-guide/sitemap/#i18n) config to the Astro sitemap integration. However, the docs seem to imply that an "alternate" link entry is created for every locale defined in this config; we need to confirm that this doesn't create entries for pages that don't exist.

  Of note: Astro's sitemap integration assumes that the lang code [is always in the first segment of the URL](https://github.com/withastro/astro/blob/main/packages/integrations/sitemap/src/generate-sitemap.ts#L35); any other placement might fail - likewise if a lang code is not provided in the config.

crutchcorn:
Closes #716

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -22,6 +22,7 @@ import mdx from "@astrojs/mdx";
 import symlink from "symlink-dir";
 import * as path from "path";
 import svgr from "vite-plugin-svgr";
+import { languages } from "./src/constants/index";
 
 await symlink(path.resolve("content"), path.resolve("public/content"));
 
@@ -36,6 +37,13 @@ export default defineConfig({
 			changefreq: ChangeFreq.DAILY,
 			priority: 0.7,
 			lastmod: new Date(),
+			i18n: {
+				defaultLocale: "en",
+				locales: Object.keys(languages).reduce((prev, key) => {
+					prev[key] = key;
+					return prev;
+				}, {}),
+			},
 			serialize({ url, ...rest }) {
 				return {
 					// remove trailing slash from sitemap URLs
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -23,6 +23,7 @@ import symlink from "symlink-dir";
 import * as path from "path";
 import svgr from "vite-plugin-svgr";
 import { languages } from "./src/constants/index";
+import { fileToOpenGraphConverter } from "./src/utils/translations";
 
 await symlink(path.resolve("content"), path.resolve("public/content"));
 
@@ -40,7 +41,7 @@ export default defineConfig({
 			i18n: {
 				defaultLocale: "en",
 				locales: Object.keys(languages).reduce((prev, key) => {
-					prev[key] = key;
+					prev[key] = fileToOpenGraphConverter(key as keyof typeof languages);
 					return prev;
 				}, {}),
 			},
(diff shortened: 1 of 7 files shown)


src/utils/translations.ts
```diff
@@ -100,12 +100,13 @@ export function getTranslatedPage(
...
 		md.file.endsWith(`${lang}.md`),
 	);
-	const enResult = globResults.find((md) => md.file.split(".")[1] === "md");
 
 	const locales = globResults
-		.map((md) => md.file.split(".")[1])
+		.map((md) => getLanguageFromFilename(md.file))
 		.filter(isLanguageKey);
 
+	const enResult = globResults[locales.findIndex((lang) => lang === "en")];
```
fennifith, line 108:
```suggestion
	const locales = globResults
		.map((md) => getLanguageFromFilename(md.file));

	const enResult = globResults[locales.findIndex((lang) => lang === "en")];
```

Looking at this, it seems like the `.filter(isLanguageKey)` could throw off `.findIndex` if it filters anything. That said, since `getLanguageFromFilename` already returns a language, I don't think the filter is needed.

astro.config.ts
```diff
@@ -36,6 +37,13 @@ export default defineConfig({
 			changefreq: ChangeFreq.DAILY,
 			priority: 0.7,
 			lastmod: new Date(),
+			i18n: {
+				defaultLocale: "en",
+				locales: Object.keys(languages).reduce((prev, key) => {
+					prev[key] = key;
```
fennifith, line 43: It isn't currently a concern (as there are no locales in `translations.json` that do this), but we may want to put these values through [`fileToOpenGraphConverter()`](https://github.com/unicorn-utterances/unicorn-utterances/blob/f2ce0dfb33b7ba66de3573f8742eb7a6d1b480d3/src/utils/translations.ts#L10) in `utils/translations` to convert from the `en-us` value (as currently described in [CONTRIBUTING.md](https://github.com/unicorn-utterances/unicorn-utterances/blob/main/CONTRIBUTING.md#finding-a-language-code)) to `en_US` as in ISO 639-1

(likewise for the meta tags)

fennifith (approved):

Issue #739, opened 2023-09-30
Accessibility: Dimmed code blocks do not have enough contrast


evelynhathaway:
On the uwu branch (PR https://github.com/unicorn-utterances/unicorn-utterances/pull/497), the code blocks content with with `.line.dim` do not mean the minimum contrast ratio of 4.5:1. The 50% opacity makes it hard to read without mousing over. This doesn't translate well to touch devices, keyboard users, or folks that do not know they have to hover the code block to make it readable.

Consider using a disclosure that has state, a visible control, and can be operated regardless of input method. Or maybe consider using another method of highlighting snippets.

Issue #738, opened 2023-09-30
Accessibility: Scrollable code blocks do not have tabindex=0


evelynhathaway:
On the uwu branch (PR https://github.com/unicorn-utterances/unicorn-utterances/pull/497), Automated tooling cannot detect the post code blocks that have horizontal scroll on `.code-container` are tabbable since JavaScript is managing focus. Using `tabindex="0"` should prevent this issue in axe-core, in case Lighthouse scores (could potentially impact search ranking) or automated tooling are important.

**Related:** These are only keyboard scrollable when JavaScript is enabled, however I don't think that it's possible to make only ones with scrollbars scrollable without JavaScript (either conditional `tabindex="0"` or focus management). If non JavaScript browsing sessions are in-scope, please consider a progressive enhancement. E.g. include `tabindex="0"` by default, and use JavaScript to remove the `tabindex` on ones that don't need scrolling.

PR #746, opened 2023-10-01
[UwU] Fix code block contrast & tabindex issues

fennifith:
- Removes the `opacity: 0.5;` styling for non-highlighted parts of code blocks
  * Instead, highlighted sections are bold + bordered
  * #739
- Fixes other code block styling issues (highlighted lines not expanding to the full code block width when scrollable)
- #738
  * Adds a rehype transform to add `tabindex="0"` on any `div.code-container`
  * Adds a script that sets `tabindex="0"` only if the `div.code-container` is horizontally scrollable

### DIFF
--- a/src/styles/shiki.scss
+++ b/src/styles/shiki.scss
@@ -63,7 +63,7 @@ Code blocks structurally look like:
 	--shiki-token-punctuation: #24292e;
 	--shiki-token-link: #24292e;
 
-	--shiki_padding: 12px;;
+	--shiki_padding: 12px;
 
 	@include darkTheme {
 		--shiki-color-text: #e1e4e8;
--- a/src/styles/shiki.scss
+++ b/src/styles/shiki.scss
@@ -104,7 +104,6 @@ pre.shiki div.line {
 	// must be the same value as used for line-height
 	min-height: 1.7em;
 	padding: 0 var(--shiki_padding);
-	border-top: 1px solid transparent;
 }
 pre.shiki div.highlight {
 	font-weight: bold;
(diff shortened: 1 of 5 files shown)


src/styles/shiki.scss
```diff
@@ -82,29 +89,34 @@ pre {
...
-}
-pre.shiki div.dim,
-pre.shiki div.highlight {
-	margin: 0;
-	padding: 0;
+pre.shiki div.line {
+	// must be the same value as used for line-height
+	min-height: 1.7em;
+	padding: 0 var(--shiki_padding);
+	border-top: 1px solid transparent;
```
evelynhathaway, line 107: How does a 2px border look to you all? Do you think it'd help with scannability?

src/utils/markdown/twoslash-tabindex/twoslash-tabindex-script.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Scrollable code-container blocks should have tabindex="0" for keyboard accessibility.
+ *
+ * These are added in rehype-transform, but can be removed with JS on elements that
+ * cannot be horizontally scrolled.
+ *
+ * https://github.com/unicorn-utterances/unicorn-utterances/issues/738
+ */
+export const enableTwoslashTabindex = () => {
```
evelynhathaway, line 9: Non-blocking: Nice! Works great when JavaScript is enabled. With it disabled, it seems like Chrome is still working around this so it behaves well, but if `tabindex="0"` was added by default, and then removed/added on load and resize, we could expand the no-JS support.

Not a crucial part of the change in my opinion.

evelynhathaway (commented):
Love it! I added two comments about potential changes. Not blocking merge

crutchcorn (approved):

src/utils/markdown/twoslash-tabindex/twoslash-tabindex-script.ts
fennifith: This was meant to be working! I thought my rehype code was doing this, but apparently not - thanks for catching that!

src/styles/shiki.scss
fennifith: Yep, that's a huge improvement IMO - I'll do that :)

evelynhathaway (approved):
Lovely lovely!

PR #785, opened 2023-10-05
[UwU] Implement redesigned heading permalink popups

### LINKED ISSUE #733: [UwU] Heading anchor links can overlap other UI elements
2023-09-28-191103_790x389_scrot

These icons should likely be moved to not overlap the table of contents.

They could also have a background/border to set them apart from other content (as they could also appear in lists and tab containers, which could also shift them around the page enough to cause overlap).

fennifith:
- Fixes #733


PrattiDev (approved):

src/views/base/scripts/heading-link-script.astro
```diff
@@ -0,0 +1,88 @@
...
+					<Icon name="link" />
+				</div>
+				 Copy link
+			</span>
+			<span data-heading-link="copied" hidden>
+				<div class={style.icon} aria-hidden>
+					<Icon name="checkmark" />
+				</div>
+				 Link copied!
+			</span>
```
crutchcorn, line 20: We need to add another span with a `aria-live="polite"` that includes the `hidden` area but not the `data-heading-link="copy"` area so that the `copied!` can be read out loud

image

crutchcorn (changes requested):
First off: WOW. This is an AMAZING implementation, especially with the `<template>` trick and all. Couldn't be happier with this implementation.

That said, I found an area where we can improve in accessibility

crutchcorn (approved):

PR #794, opened 2023-10-09
[UwU] Minor post fixes

crutchcorn:
This PR fixes tiny issues with blog posts I discovered while reviewing the UWU branch.

### DIFF
--- a/content/blog/non-decimal-numbers-in-tech/index.md
+++ b/content/blog/non-decimal-numbers-in-tech/index.md
@@ -39,7 +39,7 @@ Binary, on the other hand, is _base two_. **This means that there are only two s
 > For the Latin enthusiasts, binary comes from "binarius" meaning "two together". _Deca_, meaning 10, is where "decimal" comes from.
 > Additionally, the term "radix" is sometimes used instead of "base" when describing numeral systems, especially in programming.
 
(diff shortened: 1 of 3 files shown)


content/blog/non-decimal-numbers-in-tech/index.md
```diff
@@ -39,7 +39,7 @@ Binary, on the other hand, is _base two_. **This means that there are only two s
 > For the Latin enthusiasts, binary comes from "binarius" meaning "two together". _Deca_, meaning 10, is where "decimal" comes from.
 > Additionally, the term "radix" is sometimes used instead of "base" when describing numeral systems, especially in programming.
 
-Instead of using numbers, which can get very confusing very quickly while learning for the first time, let's use **`X`**s and **`O`**s as our two symbols for our first few examples. _An **`X`** represents if a number is present and that we should add it to the final sum; an **`O`** means that the number is not present and that we should not add it_.
+Instead of using numbers, which can get very confusing very quickly while learning for the first time, let's use **`X`** s and **`O`** s as our two symbols for our first few examples. _An **`X`** represents if a number is present and that we should add it to the final sum; an **`O`** means that the number is not present and that we should not add it_.
```
fennifith, line 42:
```suggestion
Instead of using numbers, which can get very confusing very quickly while learning for the first time, let's use **`X`**'s and **`O`**'s as our two symbols for our first few examples. _An **`X`** represents if a number is present and that we should add it to the final sum; an **`O`** means that the number is not present and that we should not add it_.
```

Suggest using apostrophes here as that'll make it read a bit better :)

fennifith (approved):

PR #820, opened 2023-10-16
[UwU] Heading order in blog post layout

### LINKED ISSUE #814: [UwU] Heading order in article pages
The article page does not currently place all of its headings under a single `<h1>` as specified in #804.

However, its current layout makes it difficult to do so:
2023-10-14-181510_351x361_scrot

The "Subscribe to our newsletter" and "Related posts" headings could be trivially moved under the post `<h1>`, but the "Table of Contents" is not in a good location for this.

We could fix this by:
- Moving the table of contents below the post heading, but retaining its visual location in the page
  The order of the page would then be "post title" (+ metadata?) -> "ToC sidebar" -> "post content"
- Creating a non-visible `<h1>` for the entire page before the table of contents, making the visible "post title" either:
  * not a heading; i.e. a `<p>` that looks like a heading
  * an `<h2>`, with the post content now starting at the `<h3>` level
- Making the table of contents heading not a heading, or deviate from the "only one top-level `<h1>`" rule

fennifith:
- Restructures the blog post layout to place all headers under the post title `<h1>`
- The `blog-post-layout` now uses `display: grid;` to accomplish this, with `"header"` as a separate slot from the main content.
  - Order in the page structure is "header" -> "sidebarLeft" -> main -> "sidebarRight"
- ~~In order to accomplish this, `<PostTitleHeader>` needed to be moved outside of `<main>` (not sure if this is the right solution)~~
  - Page structure is now: `<main><article> (article-header) (table-of-contents) (content) (related-posts) </article></main>`

Fixes #814


crutchcorn:
image

Let's add an `aria-labelled-by` on the `header <section>`

I also think we should add a hidden `h2` of `"Post contents"` and add in a `section` to `post-body` as well. That way, someone can quickly jump to contents by pressing `2` twice

fennifith:
Listing some changes to make per discussion:
- Article content should be placed in a `<section>` with `aria-describedby` an `<h2>Post content</h2>`
- The post content headings then need to start at an `<h3>` (configured in utils/markdown/index IIRC)
- Headings should use `class="text-style-headline-#"` to still visually display at their intended level (starting at `headline-2`)
- For `h5` and `h6` (in markdown, which would translate to `h7` and `h8` on the page...), use an `<h6>` with the respective styling.

crutchcorn (approved):

Issue #532, opened 2023-04-11
Nested content folder structure (suggestion)


fennifith:
The `content/posts` directory is quite large and potentially difficult to navigate at the moment - it would be a potentially useful feature for authors or collections to have their own subfolder to store particular sets of posts inside

For example...
```
content/
└── {author}/
    ├── index.md
    ├── profile.png
    ├── {post}/
    │   └── index.md
    └── {collection}/
        ├── index.md
        └── {post}/
            └── index.md
```

In this case, `{author}/index.md` would contain the current author data in the frontmatter. Posts and collections would live within a certain author folder (which does not prevent collaborative posts - although they would need to choose a primary author in the source tree).

crutchcorn:
I think I'd rather a variation on this that's easier for us to parse programatically:

```
content/
└── {author}/
    ├── index.md
    ├── profile.png
    ├── posts/{post}/
    │   └── index.md
    └── collections/{collection}/
        ├── index.md
        └── posts/{post}/
            └── index.md
```

PR #823, opened 2023-10-19
Content folder restructuring

fennifith:
This PR aims to restructure the `content/` folder following the structure in #532:
```
content/
└── {author}/
    ├── index.md
    ├── profile.png
    ├── posts/{post}/
    │   └── index.md
    └── collections/{collection}/
        ├── index.md
        └── posts/{post}/
            └── index.md
```

- Most data collections are now stored in a `Map<string, *Info[]>`
  - This stores each translation of an item in the array, mapped to its id/slug
  - This structure mostly shouldn't be interacted with directly, but through `utils/api.ts` - which also resolves the locale for each item.
- Any "*Meta" properties referencing other data types are removed/replaced with their respective queries
  - Several components now require separate `posts=` or `authors=` properties to be provided
  - The `/api/search` endpoint now returns a separate `Record<string, UnicornInfo>` rather than providing each unicorn with the post object
- `ExtendedPostInfo` and `ExtendedCollectionInfo` types are removed
  - likewise, the `rehype-unicorn-populate-post` and `rehype-unicorn-get-suggested-posts` transforms are removed
  - This mostly isolates the remark/rehype process from the rest of the post data, so it can be used separately from `Astro.glob` in a future PR.
- The `collection-mapping.ts` previously used for the FFG site has been removed/replaced with a `pageLayout: "none"` frontmatter (which just removes it from the `/collections/[slug]` paths).
- `CONTRIBUTING.md` has been updated to reflect the new file structure


content/crutchcorn/posts/advice-for-new-twitch-streamers/index.md
```diff
@@ -3,7 +3,6 @@
     title: "Advice for New Twitch Streamers",
     description: "Twitch streaming can be fun and fulfilling, but has various roadblocks in the way for newcomers. Here's how you can overcome them and grow your Twitch channel.",
     published: '2023-03-25T21:52:59.284Z',
-    authors: ['crutchcorn'],
```
crutchcorn, line 6: Was this removal intentional?

crutchcorn: Correction: It probably is, given the goals of this content folder restructure. Why don't we remove this property from every article?

crutchcorn:
I ran into a question when reading your CONTRIBUTING guide (great work BTW);

What do we do when there's two authors?

(even if there is an easy solution to this, we should probably include something in the CONTRIBUTING file)

crutchcorn (changes requested):
Something worth mentioning overall - I'm not a huge fan of how much manual mapping for data we're having to do.

I'd rather we have some kind of data mapping like we have in the codebase today.

Not inherently a blocker tho - we can revisit when we migrate away from Astro's MD processing in favor of our own.

src/views/collections/framework-field-guide/segments/quotes.astro
```diff
@@ -213,7 +213,7 @@ New software developers often don't see how programming languages share many com
 		<QuoteCard
 			personName={`Xeno`}
 			personLink="https://www.xenophorium.dev/"
-			avatarSrc={import("../../../../../public/content/data/xeno.jpg")}
+			avatarSrc={import("../people-pictures/xeno.jpg")}
```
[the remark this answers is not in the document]
fennifith: Every other profile image used in FFG is copied, which I would prefer because that means it doesn't have hidden dependencies around the repo.

If you really want to change that, could we have it resolve from `utils/api` / `getUnicornById` instead? That way it doesn't cause failures if updated in the future (e.g. these could be changed through the CMS).
(IMO this makes more sense to be static/hardcoded as it is now, though.)

content/crutchcorn/posts/advice-for-new-twitch-streamers/index.md
fennifith: This is meant to be an optional property, yes. The way that multiple authors are specified would be by listing all authors in this array, but it defaults to the author it's nested under when omitted.

src/views/collections/framework-field-guide/segments/quotes.astro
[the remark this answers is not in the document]
crutchcorn: `getUnicornById` is fine with me - let's go that route

crutchcorn:
# Testing

- [ ] About page
    - [ ] Contents
    - [ ] Translations
    - [ ] Unicorn list
- [ ] Blog post page
    - [ ] Contents
    - [ ] Multi-author
         - [ ] Social image tabs
    - [ ] Author page links
    - [ ] Social image
    - [ ] Images
    - [ ] Translations
    - [ ] Related posts
        - [ ] Authors list links
        - [ ] Title
    - [ ] Videos
    - [ ] Components
        - [ ] Tabs
        - [ ] Summary
        - [ ] iFrame
        - [ ] Table of contents
        - [ ] File tree
- [ ] Collections page
    - [ ] Table of contents
    - [ ] Social images tags
    - [ ] Epub generation
    - [ ] Author PFP
- [ ] Unicorn page
    - [ ] PFP
    - [ ] About
    - [ ] Achievements
    - [ ] Posts
         - [ ] List
         - [ ] Author links
- [ ] Search
    - [ ] Collections
    - [ ] Articles
    - [ ] Filter by author
    - [ ] Filter by tag
    - [ ] Sort
    - [ ] Show filter
    - [ ] Mobile view
    - [ ] Pagination
- [ ] Post list
    - [ ] Pagination
    - [ ] Post list (13 pages)
    - [ ] Post image cards
        - [ ] Contains in-blog-post code
    - [ ] Author list links
    - [ ] Excerpt for posts without desc
    - [ ] Desc listed
- [ ] FFG page
    - [ ] Corbin PFP/link
    - [ ] Xeno PFP/link
- [ ] Compare RSS.xml to today (needs exact match)

crutchcorn (changes requested):
image

Our About page is randomizing our contributors again

crutchcorn:
Another bug; 

All related posts have the same author as the current article:

image

image

(magic DOM is not me)

crutchcorn:
Images on the FFG site never load:

image
image
image

image

crutchcorn:
Replaced this PR with #867 due to deleted `uwu` branch

PR #840, opened 2023-10-20
Fix highlight border missing when at the last/first index in a codeblock

fennifith:
Fixes CSS for code blocks missing the top/bottom border styling when at the start or end of a code block.


crutchcorn:
This fixes the issue with the first codeblock, but seemingly not the last codeblock:

crutchcorn (changes requested):
Marking request changes as a result

crutchcorn (approved):

PR #846, opened 2023-10-21
Add Content-Security-Policy and other headers to vercel.json

fennifith:
This PR adds various headers (based loosely on [express helmet](https://github.com/helmetjs/helmet) defaults) to the Vercel config. We should make sure that none of these affect valid functionality before merging.

- [ ] Tested against FFG content (check that `alert()` works within iframes)
- [ ] Document headers in `vercel.json.md`

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -39,14 +39,25 @@
 		},
 		{
 			"source": "/(.*)",
+			"has": [
+				{
+					"type": "host",
+					"value": "unicorn-utterances.com"
+				}
+			],
+			"headers": [
+				{
+					"key": "Content-Security-Policy",
+					"value": "default-src 'self';base-uri 'self';form-action 'self' https://unicorn-utterances.com/ https://app.convertkit.com/;frame-ancestors 'none';frame-src https:;img-src 'self' https: data:;object-src 'none';script-src 'self' 'unsafe-inline';style-src 'self' 'unsafe-inline';upgrade-insecure-requests"
+				}
+			]
+		},
+		{
+			"source": "/(.*)",
 			"headers": [
 				{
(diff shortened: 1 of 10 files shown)


crutchcorn (changes requested):
This PR breaks medium-zoom with the following console error:

image

crutchcorn (changes requested):
image


The FFG site no longer smooth scrolls when the user presses "learn more" in the header as a result of this error

---------------------

Additionally, the Vercel preview URL comments bar no longer loads as a result of _this_ error:

image

crutchcorn (approved):
LGTM!

vercel.json
```diff
@@ -36,6 +36,55 @@
...
+					"key": "X-Frame-Options",
+					"value": "DENY"
+				},
+				{
+					"key": "X-Permitted-Cross-Domain-Policies",
+					"value": "DENY"
+				},
+				{
+					"key": "X-XSS-Protection",
+					"value": "0"
```
evelynhathaway, line 85: Curious to know why we'd disable XSS protection

evelynhathaway (changes requested):
I cannot sign into Vercel preview comments

image

vercel.json
fennifith: [According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection): "Even though this feature can protect users of older web browsers that don't yet support CSP, in some cases, XSS protection can create XSS vulnerabilities in otherwise safe websites." There are a variety of recommendations to use `X-XSS-Protection: 0` because of this - which is also in [helmet.js](https://github.com/helmetjs/helmet#x-xss-protection) and most major sites.

PR #865, opened 2023-10-28
fix: table of contents needs proper bottom padding

### LINKED ISSUE #864: Table of contents needs proper bottom padding
We need to apply a bottom padding of var(--site-spacing) to the table of contents. This is to make sure it is abundantly clear that it is the end of the list, as opposed to the situation below, where it's enough space to disable scrolling, but not enough to show whether or not there's another entry out of sight. 

image

maheshchandra10:
Fixes #864


crutchcorn (approved):
Thank you so much for your contributions!

I've mentioned it before, but we'd love to have ya in our Discord if you'd be interested - there's a `developer` role waiting for ya if you'd like it :) 

https://discord.gg/FMcvc6T

PR #940, opened 2023-12-14
chore: fix PT_BR locale filename and language

crutchcorn:
This seems to've removed our sitemap somehow. Let's see if this fixes it


crutchcorn:
Sure enough, that fixed it

PR #1015, opened 2024-01-20
[Blog Post] New to Web Development? Don't Use React.

xenophorium:


crutchcorn (approved):
I may not personally agree with everything in this article, but that's what the opinion pieces are for! 

Well written, approved on my end :)

PR #1025, opened 2024-02-12
Add seo preload header for font so it loads earlier in rendering

bzbetty:


crutchcorn:
CCing @fennifith for review

BTW @bzbetty do you have a Discord? We'd love to have you in our community and give you a `Developer` role!

https://discord.gg/FMcvc6T

Also, feel free to add yourself to `unicorns.json` and add "developer" in the role array to have yourself on our about page

src/styles/text-styles.scss
```diff
@@ -44,6 +44,7 @@
 	font-size: var(--h1_font-size);
 	font-weight: var(--weight_black);
 	line-height: var(--h1_line-height);
+	font-display: swap;
```
fennifith, line 47: This would be a good change! Should the same preload header also be added for the other font-faces?

I'm curious about this `font-display` - I think the `fallback` option might be better to avoid layout shift when the font loads? (this is a bit of a middle ground between appearance vs. speed, though)

It also looks like `font-display` needs to go on the `@font-face` rule, not on the CSS class that uses it ;)

https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display

bzbetty: yeah fallback might be better for layout shift, depends on how different the fonts are. to be honest it's probably not really needed on this site, and as you pointed out I seem to have gotten it wrong anyway.

Might be worth just trying the preload by itself? main goal was to fix the LCP delay which is why it was only for that one font face.

bzbetty: I removed the swap for now as I didn't have the time to ensure it was working right.  The preload worked great though and it since it was mainly for the LCP i didn't see the need to add it for the other fonts.

I'd say get this small change merged in then worry about other potential changes when you see the google metrics next.

fennifith: Sounds good to me! Thank you for the PR :D

fennifith (approved):

Issue #850, opened 2023-10-21
Change page meta descriptions for each page number


crutchcorn:
<img width="814" alt="" src="https://github.com/unicorn-utterances/unicorn-utterances/assets/9100169/7bca4051-7236-42fc-b161-4ebfb0ecb1ea">

fennifith:
Closing this as it was fixed in #1029 :)

PR #1029, opened 2024-02-17
Fix: Change page meta descriptions for each page number #850

rkarlovic:
Added SEODescription so we can use it for description in <SEO />. I have used existing description but added  at end`Browse all posts on Unicorn Utterances, page ${pageIndex}``.

image

image


src/pages/page/[page].astro
```diff
@@ -17,10 +17,12 @@ const { page } = Astro.props as { page: Page<PostInfo> };
 const pageIndex = page.currentPage;
 
 const SEOTitle = `Post page ${pageIndex}`;
+
+const SEODescription = `Learning programming from magically majestic words. A place to learn about all sorts of programming topics from entry-level concepts to advanced abstractions. Browse all posts on Unicorn Utterances, page ${pageIndex}`;
```
Comment (crutchcorn) [inline src/pages/page/[page].astro:21]: Let's change this to:

```suggestion
const SEODescription = `Browse all posts on Unicorn Utterances, page ${pageIndex}`;
```

fennifith (approved):

Issue #972, opened 2023-12-27
[Website Bug]: inconsistent font size across code blocks


jocdiazm:
### Description

Hello team ,

I’m fairly new to the website content but wanted to report a possible bug regarding font size when navigating code blocks.

### Minimal Reproduction

1. Use a browser in iOS (happens on safari and chrome)
2. Go to a post and scroll to find code blocks
3. Notice the different font sizes across them

Example [post: what is use form state and status](https://unicorn-utterances.com/posts/what-is-use-form-state-and-status)

screenshots: 

![IMG_8328](https://github.com/unicorn-utterances/unicorn-utterances/assets/13368066/71dfff4c-15cc-4679-b0ad-b818d67e2540)
![IMG_8327](https://github.com/unicorn-utterances/unicorn-utterances/assets/13368066/e486c94f-54de-436c-ac04-97e6d796af77)

### Exception or Error

None

crutchcorn:
Thanks for reporting this! I'll have to take a look sometime in the new year to figure out what could be going on

PR #1035, opened 2024-02-20
Fix(#972): different code font-size on iOS

MateuszWasik:
Closes [#972](https://github.com/unicorn-utterances/unicorn-utterances/issues/972)

On iOS devices (Safari, Chrome) code font-size inside markdown files was bigger.

Before: 
before

After:
after

### DIFF
--- a/src/styles/post-body.scss
+++ b/src/styles/post-body.scss
@@ -97,6 +97,8 @@
 	pre,
 	code {
 		@extend .text-style-code;
+		// Ensures consistent font-size within code blocks on iOS devices
+		-webkit-text-size-adjust: 100%;
 	}
 
 	code:not(pre code) {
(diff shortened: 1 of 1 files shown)


src/styles/post-body.scss
```diff
@@ -165,7 +167,10 @@
 		//
 		// - this prevents elements targeted by the URL hash (e.g. "/some-page#Heading-ID")
 		//   from being obscured by the header
-		scroll-margin-top: calc(var(--header_logo_size) + var(--header_padding-vertical)*2 + 1px + var(--site-spacing));
+		scroll-margin-top: calc(
```
MateuszWasik, line 170: Prettier decided to reformat that as well :)

crutchcorn (approved):
I've confirmed this fixes the issue on iOS and:

- Doesn't change behavior on Firefox or Chrome (Desktop)
- Doesn't break font scaling on iOS
- Doesn't break font scaling on Firefox or Chrome (Desktop)

Great work!

PR #1049, opened 2024-03-12
Removes duplicate "Custom hooks" in TagList.

mark-keaton:
Custom hooks is listed twice in the TagList which results in it showing twice on the home page:

image

I'm hoping this will resolve that.


crutchcorn:
Unfortunately, there's not much I can do to fix this. We need 30 elements in the array for the math to work out with repeating items.

This is why tsc is failing; I enforce the array length at build time.

That said; thanks a bunch for opening this Mark! You definitely didn't have to but it's immensely appreciated 😊

PR #1055, opened 2024-03-14
feat: rehype plugin expand details and summary before rendering to epub

MateuszWasik:
[Closes 1051](https://github.com/unicorn-utterances/unicorn-utterances/issues/1051)

Adding new rehype plugin that will expand `<details>` before rendering to ePub

- transform `<details>` and `<summary>` into `divs` 
- apply `chevron down` unicode element to `<summary>` element as first children

image


MateuszWasik:
Right now plugin is connected to `html` instead of `epub` just for testing purposes. `<details>` section is being expanded as expected.

src/utils/markdown/index.ts
```diff
@@ -108,6 +109,7 @@ export function createRehypePlugins(config: MarkdownConfig): RehypePlugin[] {
 					rehypeUnicornElementMap,
 					rehypeTwoslashTabindex,
 					rehypeFileTree,
+					rehypeExpandDetailsAndSummary,
```
crutchcorn, line 112: Let's move this to the EPUB format, remove the `[WIP]` title, then we can merge

crutchcorn (changes requested):
Minor change requested. Otherwise LGTM! :D

src/utils/markdown/index.ts
MateuszWasik: It's in HTML format for a testing purposes. Unfortunately when it's moved to EPUB format it's not working as expected.

I tested it on macOS, ipadOS and on Kindle ebook reader. In those cases <details> elements are not expanded, even though this rehype plugin works correctly on HTML format.

MateuszWasik: I can't find a solution for this behaviour.

MateuszWasik: As discussed, we need to transform those elements into `divs` and add a new class that will style them to look like `<details>` elements

MateuszWasik:
So actually `svg` elements is a miss or hit as @rjt-rockx says :D He actually proposed that we can use unicode to display `chevron` and that actually was a great idea. 

@crutchcorn @fennifith what do you think? Do we miss something by using unicode instead of svg element here?

src/utils/markdown/rehype-expand-details-summary.ts
```diff
@@ -0,0 +1,43 @@
+import { visit } from "unist-util-visit";
+import { Element, Root } from "hast";
+
+const unicodeElement: Element = {
+	type: "element",
+	tagName: "span",
+	properties: {
+		style: "margin-right: 3px;",
```
MateuszWasik, line 8: We have to add some margin because `chevron` and `title` are too close to each other.

crutchcorn (approved):
Good work!

Issue #1041, opened 2024-02-29
[Website Bug]: Cancelling search results removes controls & causes back navigation bug


meows:
### Description

When you visit this search experience:

https://unicorn-utterances.com/search?searchQuery=react&filterAuthors=crutchcorn&filterTags=angular

- And then you `(X)` the search result, the filters will disappear from screen and there's no obvious way to get them back.
- Pressing back from this state changes the URL but won't do anything for a few tries.
- Typing the same search query will remove the filters.

fennifith:
Opened #1061 for the navigation bug. The filters disappearing is currently the intended behavior, as described in #861 - feel free to comment there if you have any thoughts on it :)

PR #1061, opened 2024-03-17
Improve search page params handling

fennifith:
- All search parameters are deserialized in one place into the `SearchQuery` type - which simplifies a lot of the `useMemo` logic in `search-page.tsx`. This also consolidates the default values for these params.

- Fixes the navigation bug mentioned in #1041

- The search query URL is now debounced separately from the search query.
  * This reduces the search debounce delay to the one 500ms `useDebouncedValue`, and allows the UI to re-render immediately without waiting another 500ms for the query URL to change

- The pagination component now passes the page number alongside the href to the `softNavigate` callback (which avoids unnecessarily parsing out the page number)


MateuszWasik:
LGTM 👍

MateuszWasik (approved):

Issue #1067, opened 2024-03-21
`<iframe>` with `uu-code` links need to be removed entirely


MateuszWasik:

fennifith:
We've decided to replace this with different behavior as detailed in #1027

PR #1068, opened 2024-03-24
feat: remove iframe with uu code url from epub

MateuszWasik:
[Closes](https://github.com/unicorn-utterances/unicorn-utterances/issues/1067)

- Remove iframe elements from EPUBs whose URLs include a UU code.

Prepared two `iframe` components that includes `uu-code` and two that isn't. 

Screenshot 2024-03-24 at 15 01 45

This is how the EPUB output look like.

image


MateuszWasik:
@fennifith so as discussed on discord channel, we no longer have to remove the `uu-code` iframes. So this PR and according task in `Issues` can be closed?

fennifith:
Yep! I'll update the issue details in a sec :)

PR #1078, opened 2024-04-15
docs: fix Remix and Vercel links in FFG's preface

alejandrocuba:
This PR fix the incorrect links to Remix and Vercel in the Preface


crutchcorn (approved):
Thanks for this!

Issue #1072, opened 2024-04-05
[Website Bug]: "View This Post on Github" link broken (in some cases?)


joezimjs:
### Description

I was on https://unicorn-utterances.com/posts/ffg-fundamentals-accessing-children and clicked the link at the bottom of the page to "View This Post on Github" and it took me to https://github.com/unicorn-utterances/unicorn-utterances/tree/main/content/blog/ffg-fundamentals-accessing-children/index.md which was a 404

### Exception or Error

404

alejandrocuba:
This issue is also reproducible in other blog posts across the site, not just in FFG content.

PR #1079, opened 2024-04-15
Fix GitHub link on posts

rusher2004:
Fixes the issue in #1072, where you'd land on a 404 when clicking to view the post on GitHub.

The link was not updated from when the content directory was restructured. Now we can just use the post path to construct the url.

Screen Recording 2024-04-15 at 9 41 24 AM


fennifith (approved):
Awesome fix - thank you for the PR! :D

Issue #1076, opened 2024-04-12
[Website Bug]: Layout shift when switching between the React/Angular/Vue code tab in Safari 


bk7312:
### Description

Layout shift when switching between the React/Angular/Vue code tabs in Safari (or any browser on iOS and iPadOS). Does not happen on other desktop/Android browsers.

### Minimal Reproduction

1. In Safari, open Part 2 of the Framework Field Guide, https://unicorn-utterances.com/posts/ffg-fundamentals-intro-to-components, and scroll to the code tabs.
2. Switch between the React/Angular/Vue tabs.
3. The code tab will jump to the vertical center of the page.
4. Scroll up or down a bit and repeat step 2 to repeat the behavior.

### Exception or Error

NA

fennifith:
Thanks for the report!

This was the intended behavior - as a workaround since Safari doesn't support [CSS overflow-anchor](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor) just yet. That said, I'll leave this open as there might be a better JS solution that preserves the scroll position.

For anyone looking at contributing: [`src/utils/markdown/tabs/tabs-script.ts`](https://github.com/unicorn-utterances/unicorn-utterances/blob/f24b6eb700932c50298090d973af02e866ad8efd/src/utils/markdown/tabs/tabs-script.ts#L67) is the place to look :)

Gungier:
@fennifith I can see if I can implement something similar as the last fix

bk7312:
@fennifith @Gungier I've created a PR to solve this issue, please have a look. Thanks

PR #1082, opened 2024-04-18
Fix safari layout shift bug (#1076)

bk7312:
Fixes issue #1076. Instead of using `target.scrollIntoView()`, calculates the offset before running `changeTabs()` and apply that offset to the window vertical scroll position using `window.scroll()`. This ensures the position of the code tabs remain the same before/after changing tabs.


src/utils/markdown/tabs/tabs-script.ts
```diff
@@ -45,15 +45,15 @@ export const enableTabs = () => {
...
-			}, 0);
+			// Redundant: Scroll handled by `handleClick` on `tab.click()`
+			// setTimeout(() => {
+			// 	tab.scrollIntoView &&
+			// 		tab.scrollIntoView({
+			// 			behavior: "auto",
+			// 			block: "center",
+			// 			inline: "center",
+			// 		});
+			// }, 0);
```
fennifith, line 56: This is still needed because keyboard interaction will not necessarily keep the tabs in view; we should ensure that the buttons are visible here even if `shouldScrollToTab` is false.

Unfortunately, we can't rely on `tab.focus()` or the `block: "nearest"` option for `scrollIntoView()`, as it does not take the page's sticky header into account. The tab buttons can still be focused and concealed behind the header, because the browser only checks that they are within the viewport.

If you're up for it, I do think this could be improved with the following:
- Replace`tab.click()` with a `changeTabs(tabName)` call (so that it never redundantly runs the `window.scroll` in `handleClick`)
- Uncomment the following `setTimeout` / `tab.scrollIntoView`
- Add a comment to document why this is done (tbh I should've done that last time I touched this, so this is my bad)

fennifith (commented):
Hey, thanks for the PR!

I haven't tested this yet (don't have a device with Safari) but the fix itself looks great! Just one issue with `handeKeydown`:

src/utils/markdown/tabs/tabs-script.ts
bk7312: I've replaced `tab.click()` with `changeTabs(tabName)` and also replaced the deprecated `e.keyCode` with `e.code`. 

For the scroll, I've added some code to preserve the scroll location if the tabs are within the viewport, else it will scroll to 20px below the header. Let me know if this is preferred, might also be worth implementing this on the table of contents so it scrolls to 20px below the header instead of center?

Gungier: @fennifith @bk7312 I can test this on my mac when I get a moment later.

crutchcorn (approved):
LGTM! Thanks for this! Sorry for the long wait

PR #1122, opened 2024-08-05
New "Join Us" page - Canvas implementation

fennifith:
Ports the contributors header to a canvas implementation - this significantly reduces lag on mobile devices when the bubbles are animating.

Also makes a few behavior changes:
- Uses `getPicture` to obtain sized image URLs from Vercel (passed in through `data-unicorns` on the canvas)
- Makes all animation timestamp-dependent to avoid behavior differences due to refresh rate
- Uses [passive event listeners](https://developer.chrome.com/docs/lighthouse/best-practices/uses-passive-event-listeners/) that won't block scrolling on mobile
- Prevents vertical drags (i.e. scrolling on mobile) from moving the canvas

### DIFF
--- a/src/views/join-us/contributor-card.astro
+++ b/src/views/join-us/contributor-card.astro
@@ -43,7 +43,7 @@ const isPartner = unicorn.achievements.includes("partner");
 			{
 				isPartner && (
 					<span class={`text-style-body-small-bold ${style.partnerLabel}`}>
-						<img src="/flair/bowtie.svg" alt="" />
+						<img width="16" height="16" src="/flair/bowtie.svg" alt="" />
 						Partner
 					</span>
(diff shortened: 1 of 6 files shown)


crutchcorn (commented):
This is awesome. One question tho - the script loads pretty slow on mobile and such, meaning that there's a massive blank space until the script loads (which is `defer` because the script is `module` right now from Astro)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#other_differences_between_modules_and_standard_scripts

Can we add `is:inline` to it so that we can sidestep this problem?

PrattiDev:
Looks good to me!

crutchcorn (approved):

PR #1131, opened 2024-08-22
feat: orama search

crutchcorn:
This PR replaces our in-house Fuse.js powered search with one from [Orama](https://askorama.ai/) to make our results much better.

#sponsored (disclosure to follow soon)

# TODOs

- [x] Add author name/handles (object arrays not supported in indexes yet in Orama)
- [x] Add titles of chapters in collections array (`string[]`)
- [x] `*` search for all results

## Deployment

- [ ] Merge this PR
- [ ] Update the Remote JSON urls in Orama to point to prod
- [x] Update the blog post date


src/views/search/components/search-footer.tsx
```diff
@@ -0,0 +1,41 @@
...
+		setIsDarkTheme(
+			document.documentElement.classList.contains("dark")
+		);
+	};
+
+	useEffect(() => {
+		const observer = new MutationObserver(handleMutation);
+		observer.observe(document.documentElement, { attributes: true });
+		return () => observer.disconnect();
+	}, []);
```
crutchcorn, line 23: Move `handleMutation` inside of `useEffect` to avoid issues with linting or stale data

src/views/search/orama.tsx
```diff
@@ -0,0 +1,189 @@
...
+export const SearchClient = createContext<SearchContext>(undefined as never);
+
+interface OramaClientProviderProps extends PropsWithChildren {
+	params?: Partial<ConstructorParameters<typeof OramaClient>[0]>;
+}
+
+export function OramaClientProvider(props: OramaClientProviderProps) {
+	const postClient = new OramaClient({
+		endpoint: "https://cloud.orama.run/v1/indexes/playful-programming-p9lpvl",
+		api_key: "OLeHrFPWLR0alSSZkMiq4tokMZZNEbDL",
```
crutchcorn, line 49: Let's move the `api_key` and `endpoint` into a `constants` file to keep track of it better

src/views/search/orama.tsx
```diff
@@ -0,0 +1,189 @@
...
+function addMerge(obj1: Record<string, number>, obj2: Record<string, number>): Record<string, number> {
+	return Object.entries(obj2).reduce(
+		(acc, [key, value]) => ({ ...acc, [key]: (acc[key] || 0) + value }),
+		{ ...obj1 },
+	);
+}
+
+export async function searchForTerm({ postClient, collectionClient }: SearchContext, query: SearchQuery, signal: AbortSignal) {
+	// Schema should be passed to `search` method when:
+	// https://github.com/askorama/oramacloud-client-javascript/pull/35
```
crutchcorn, line 89: This was merged. We can change this now

crutchcorn (commented):
@fennifith suggested a few small changes. We also need to add Orama to our sponsors list here:

https://playfulprogramming-git-orama-search-oceanbit.vercel.app/join-us

src/views/search/orama.tsx
fennifith: This doesn't appear to be working. I'm only seeing this function in the type definitions, and not the version with generics (using the latest version):
```ts
search(query: ClientSearchParams, config?: SearchConfig): Promise<Nullable<Results<AnyDocument>>>;
```

I'll update this comment & the client lib version, but will look into what's happening with the types later. (I don't think it should block this PR)

PR #1150, opened 2024-09-26
ToC fix, new Bluesky link, new Twitter icon

PrattiDev:


content/data/about.json
```diff
@@ -4,18 +4,22 @@
 			"icon": "discord",
 			"url": "https://discord.gg/FMcvc6T"
 		},
+		"(Twitter)": {
```
crutchcorn, line 7:
```suggestion
		"X (Twitter)": {
```

crutchcorn (changes requested):

PR #1216, opened 2024-11-30
Add overflow-wrap:break-word to enforce text wrapping

### LINKED ISSUE #1213: Long links in articles break layout by not going to the next line
### Description

Long links in articles break layout by not going to the next line.

Links should have appropriate CSS: `word-break: break-word;`

### Minimal Reproduction

Visit this link on a mobile device or with Browser Developer Tools set to Mobile: https://playfulprogramming.com/posts/modern-js-bundleless

Observe the page width.

### Exception or Error

_No response_

fennifith:
Fixes #1213

Used [`overflow-wrap: break-word;`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) in place of `anywhere` as it does not affect content size calculation:
> soft wrap opportunities introduced by the word break are NOT considered when calculating min-content intrinsic sizes.

### DIFF
--- a/src/styles/text-styles.scss
+++ b/src/styles/text-styles.scss
@@ -37,6 +37,12 @@
 
 :root {
 	font-family: var(--pfp-font-family);
+
+	// Prevent long text from overflowing if there are no acceptable break points
+	// (this is necessary as Chrome/etc have different text wrap behavior)
+	// https://github.com/playfulprogramming/playfulprogramming/issues/1213
+	// https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap
+	overflow-wrap: break-word;
 }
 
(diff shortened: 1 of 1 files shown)


crutchcorn:
Not opposed to merging this as-is, but thoughts on using the `/` `<wbr>` parsing instead of this trick? I think that might end up with a cleaner result for most things

fennifith:
I feel like both would be useful - e.g. if someone types out a very long inline code snippet, we'd still want it to wrap rather than creating horizontal overflow even if it doesn't have any sensible breakpoints, and I couldn't really find any downsides to it.

I'm less enthusiastic about writing a rehype plugin for `<wbr>` as there are so many different edge cases, though.

PR #1242, opened 2024-12-12
[Blog Post] Mastering Angular Unit Testing: Best Practices and Tools

sonukapoor:


PrattiDev:
First of all, great article. Very comprehensive!

From my quick review, there's a typo in this line: (`[]` are fixes)

```
mocking classes that [are] several layer[s] down the dependency tree.
```

In your **Setting providers** chapter, you link to documentation, but the link is specifically tagging `#providers` as opposed to the `#componentproviders`. To make matters worse, in the documentation, the linkback to contentProviders is broken because it's not all lowercase.

I'll leave it at your discretion to expand it but I do think we should at least explain when to use `providers` vs `componentProviders` if the link was intentional.

sonukapoor:
@PrattiDev Thank you. I will address the issues in the next few days.

sonukapoor:
@PrattiDev Fixed the typos. 

The link was meant to point the user to the provider section, which is what I am doing.

PrattiDev:
@sonukapoor It's all good! We'll be publishing it this week. Thank you so much for the contribution!

PR #1260, opened 2025-01-20
Fix collection card sizing on small viewports

### LINKED ISSUE #1255: [Website Bug] Collection card button does not size correctly on small viewports
Image

This is caused by flexbox when the collection authors wrap to multiple lines. Also note that the cover image has empty space at the bottom, which should not be there.

fennifith:
Fixes #1255 

Adds `align-self` so that the collection card button & cover image do not automatically use the max size of the flex container.


crutchcorn:
Unrelated, but I noticed:

image

Do we wanna fix that in this PR or a different?

PR #1282, opened 2025-02-01
[Blog Post] Why Can't Angular Cast An Element to Another?

crutchcorn:
> This blog post references the ones from #1287 and #1290, please release those first.

### DIFF
--- a/content/crutchcorn/posts/angular-why-no-template-casting/index.md
+++ b/content/crutchcorn/posts/angular-why-no-template-casting/index.md
@@ -1,0 +1,164 @@
+---
+{
+	title: "Why Can't Angular Cast An Element to Another?",
+	description: "",
+	published: '2025-12-01T13:45:00.284Z',
+	tags: ['angular', 'webdev', 'javascript'],
+	license: 'cc-by-nc-sa-4'
+}
+---
+
+While I'm a huge fan of Angular at heart, I've often used React at my day jobs.  While working with React component libraries - either internal or external - you're likely to run into a pattern like so:
+
+```jsx
+<OurButton as="a" href="oceanbit.dev">This looks like a button, but is a link</OurButton>
+```
+
+This `<OurButton/>` component is able to expose an internal `"button"` tag when nothing is passed, but transform into any other `as` element when the property is passed.
+
+What's cooler is that the other attributes from the `as` original element (like `<a>`'s `href` above) can be type-safe using some TypeScript magic.
+
+You can implement it in Vue, too!
+
+<!-- ::start:tabs -->
+
+# React
+
+```tsx
+import {
+	ComponentPropsWithoutRef,
+	ElementType,
+	PropsWithChildren,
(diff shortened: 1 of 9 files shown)


content/crutchcorn/posts/angular-why-no-template-casting/index.md
```diff
@@ -0,0 +1,174 @@
...
+
+This is all to say _"Angular has minimal control over how to create or update original DOM structures once it's hit the browser"_. Once it's past the compilation stage, **the most Angular is capable of doing is creating a single DOM node that's not part of the existing template blueprint function** from the compiler output.
+
+> This is partially how Angular is able to keep such slim and expedient runtimes; it has lower required overhead because it doesn't need to handle arbitrary DOM manipulation and can rely on persistent blueprints of your template to execute as needed.
+>
+> This is also why the Angular team has been hesitant to roll out JSX as an authoring format for Angular; it's not a good fit as JSX's templating is intended to allow for dynamic transformation at runtime.
+
+In order for there to be some kind of `as` casting API akin to an imaginary API of `*ngElementNameOutlet="tag"`, we'd need a more expressive runtime component to Angular; which would instantly lead to heavy bloat in our bundles, extended initial JavaScript parsing time, and more.
+
```
ajitzero, line 174: suggestion: I know this is in a draft state (already very well-written!) but would you consider giving an example at the end for the Angular alternative/solution of directive-based selectors?

e.g. Angular Material has [mat-button](https://material.angular.io/components/button/overview#button-overview) which can be used on either a `<button>` tag or an `<a>` tag. In this use-case, the authoring experience becomes:

```diff
-<MatButton as="button" disabled>Disabled</MatButton>
+<button mat-button disabled>Disabled</button>
-<MatButton as="a" href="#" target="_blank">Link</MatButton>
+<a mat-button href="#" target="_blank">Link</a>
```

IMO, this removes the need for props spreading (<Comp {...props}>) since, with directive selectors, we have direct access to the main DOM element. With React/Vue, the component author needs to support those inputs and pass them on to the actual element. Not _really_ a pro or con, but the article so far (which is still a draft and maybe you thought of this already) makes it look like a limitation of Angular.

crutchcorn: I may touch on this, this is what Alex (from the Angular team) suggested as well as an alternative. 

The only challenge I have to what you've outlined here is that this _is_ a limitation of Angular. There are good reasons for it, and it can be worked around, of course, but from a design system building perspective it does require a different API.

This article was actually born by finding this limitation (again) while writing something unrelated.

crutchcorn: BTW! Are you in our Discord? We'd love to have you there: https://discord.gg/FMcvc6T

ajitzero: Joined the Discord!

Issue #1256, opened 2025-01-20
Enable search Hybrid mode


fennifith:
See: #1196 

Currently, pagination is not supported in hybrid Orama queries. While we _could_ grab the maximum amount of results and paginate in UI-only, the response limit means that hybrid queries would not return every possible result on the site; rather, only the results that are most relevant.

There are a few options to enable this:
- A toggle to intentionally use a hybrid mode search
    * I don't think this is worth the cognitive load it adds, UX-wise
- Automatically switch to hybrid mode in certain conditions
    * e.g. use fulltext for `*` searches and short queries, but hybrid for longer queries that might have fewer results

crutchcorn:
@ckhackshaw is going to work on this ticket per a convo on our Discord

ckhackshaw:
@crutchcorn, I have a PR ready for this. See https://github.com/playfulprogramming/playfulprogramming/pull/1285

The contributing guide covers post contributions so I am not familiar with the process used at PP for merging. Let me know if I missed any additional steps that I need to do.

Maybe something to look at.

ckhackshaw:
I think this issue can be closed now

@crutchcorn @fennifith

crutchcorn:
Good call!

PR #1285, opened 2025-02-02
Enable Orama hybrid search in some searches

ckhackshaw:
This PR adds functionality to enable hybrid search when a user types 3 or more words in the search bar. It deals with issue [#1256](https://github.com/playfulprogramming/playfulprogramming/issues/1256)

### DIFF
--- a/src/views/search/orama.tsx
+++ b/src/views/search/orama.tsx
@@ -6,7 +6,13 @@ import { PropsWithChildren } from "components/types";
 import { createContext } from "preact";
 import { useContext } from "preact/hooks";
 import { SearchQuery } from "./search";
-import { ORAMA_COLLECTIONS_API_KEY, ORAMA_COLLECTIONS_ENDPOINT, ORAMA_POSTS_API_KEY, ORAMA_POSTS_ENDPOINT } from "./constants";
+import {
+	ORAMA_COLLECTIONS_API_KEY,
+	ORAMA_COLLECTIONS_ENDPOINT,
+	ORAMA_POSTS_API_KEY,
+	ORAMA_POSTS_ENDPOINT,
+	ORAMA_HYBRID_SEARCH_ACTIVATION_THRESHOLD,
+} from "./constants";
 
 const postSchema = {
 	slug: "string",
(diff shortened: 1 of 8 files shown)


src/views/search/orama.tsx
```diff
@@ -69,27 +75,47 @@ export function OramaClientProvider(props: OramaClientProviderProps) {
...
+	const sortBy: SortByClauseUnion | undefined =
+		query.sort === "relevance"
+			? // When term is empty (returning all results), there is no "relevance" to sort by - so this defaults to a sort by newest
+				term.length > 0
+				? undefined
+				: { property: "publishedTimestamp", order: "desc" }
+			: {
+					property: "publishedTimestamp",
+					order: query.sort === "newest" ? "desc" : "asc",
+				};
```
crutchcorn, line 118: Nit: Let's not use this many nested turnaries. Instead, let's:

```javascript
let sortBy: SortByClauseUnion | undefined;

// et al
if (something) {}
```

src/views/search/search-page.tsx
```diff
@@ -158,8 +172,25 @@ export function SearchPageBase() {
...
+	const authorCounts = usePersistedEmptyRef(
+		isWildcardSearch
+			? Object.fromEntries(
+					people.people.map((person) => [person.id, person.totalPostCount]),
+				)
+			: data.authors,
+	);
+
+	// if searchh term has more than a certain number of words, then use hybrid mode Orama search for smart/AI searching capabilities
+	const isHybridSearch =
```
crutchcorn, line 191: This seems like it could use a `useMemo` for perf reasons to avoid re-running every render

crutchcorn (changes requested):
Some comments here. Otherwise, this is awesome! Huge work!

Also, something I noticed: `Sort: ` doesn't work as intended. Can you investigate and see if that's a bug with us or with Orama?

src/views/search/search-page.tsx
ckhackshaw: Will do.

ckhackshaw:
Should I tackle this under a new issue? 

Asking since I didn't touch anything related to sorting in these code changes.

It might make sense to roll this problem along with e refactoring you suggested re: ternaries into a separate issue with a new PR.

Thoughts?

src/views/search/orama.tsx
ckhackshaw: See comment regarding sorting bug

crutchcorn:
Let's do the ternary refactor + useMemo here, but the sorting issue in a new PR.

src/views/search/search-page.tsx
ckhackshaw: @crutchcorn, completed. Let me know if this is ok.

src/views/search/orama.tsx
ckhackshaw: @crutchcorn, refactor done. I've put the sort bug as a todo to complete soon.

Issue #1271, opened 2025-01-25
Article revision feature


crutchcorn:
- [x] Implement design + dropdown
- [x] Make dropdown accessible
- [x] Hook into articles for real
    - Use `version` to show the version name
    - Sort based off of `published` date

# Preview

<img width="685" alt="Image" src="https://github.com/user-attachments/assets/c9a3c980-3034-41ff-8d9b-9676edb3f134" />

# Link to Figma

https://www.figma.com/design/Ul8KB1Sc6etCNTRCRSc27z/Playful-Programming---Website?node-id=8196-41037&t=OJatibnQ2ZBjE5lC-4

ckhackshaw:
I can work on this. Can you assign this to me @crutchcorn ?

crutchcorn:
SGTM! Feel free to work on in @ckhackshaw

crutchcorn:
WIP work in #1299

crutchcorn:
Blocked until #1160 is merged

PR #1299, opened 2025-02-10
Article revision feature

ckhackshaw:
@crutchcorn 

This is a very basic WIP of the direction I am looking at with this feature.

As you can see, I am looking to make use of the Select component already in use in the project. Reason being, it has all the functionality that I need so make no sense reinventing the wheel. 

I will just need to update the design, so I am thinking, an option to override the default styles but obviously keep default styles too.

1. What do you think?

I have one issue right now, the dropdown isn't working. I will investigate but feel free to lmk if there's anything that you see that sticks out to you that I haven't done/forgot which is causing/contributing to that issue. I haven't added any state or state setters as yet but I don't think that should affect the functionality.


ckhackshaw:
Fixes #1271

crutchcorn (changes requested):
We cannot use the existing `Select` component, as it uses Preact and some other heavy libraries that we do not want in our blog post pages (too large of a bundle size).

Instead, we should use the new Popover API:

https://developer.mozilla.org/en-US/docs/Web/API/Popover_API

To build out a different select component without Preact.

You can find a similar demo here: https://mdn.github.io/dom-examples/popover-api/nested-popovers/ (View source) to see how it works

fennifith:
Summarizing some details from our discussion :)

Select popup component / behavior:
- Using Preact on the post pages is perfect, but the existing Select component still shouldn't be used as it relies on `react-aria` (which is a much bigger dependency)
- Using the `popover="auto"` mode should enable most of the popup behavior without JS (onClick isn't needed!)
- The popover positioning should be handled with `position: relative;` on the parent container, and `position: absolute;` on the popup itself.
  - This way `top:` and `left:` properties are treated as an offset from the parent, and the element scrolls with the page without needing a scroll listener.

Getting the list of post versions:
- `version: string` and `upToDateSlug: string` can be added as properties to the `RawPostInfo` interface
- These can be accessed from the posts map in [data.ts](https://github.com/playfulprogramming/playfulprogramming/blob/e0da0ceda3ed477e1483a8c5538d6c4dc82e9aac/src/utils/data.ts#L361)
- To get a list of versions from a post, I'd add a `getPostVersionsBySlug(slug: string): PostInfo[]` function to [api.ts](https://github.com/playfulprogramming/playfulprogramming/blob/e0da0ceda3ed477e1483a8c5538d6c4dc82e9aac/src/utils/api.ts)
  - This can scan through the posts map for any entries with a matching `upToDateSlug` and sort those entries by date

ckhackshaw:
@fennifith have a look at the most recent work done, few things to note:

- With relation to the Popover API, I have used native html features, so no JS as was initially requested
- Relative and absolute positioning doesn't work here, however we can use [CSS anchor positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning), which I did. See browser support [here](https://developer.mozilla.org/en-US/docs/Web/CSS/position-anchor#browser_compatibility)
- Also, Popovers are not in normal document flow, they are in the top layer, so the menu is above the navigation bar when one scrolls down the page with the menu open. I tried `position-visibility` [see here](https://developer.mozilla.org/en-US/docs/Web/CSS/position-visibility) but again, has low support

If you are happy with these constraints, let me know so that I can convert `ArticleRevisionDropdown` to an astro component

1. I used the publishedMeta property for the dates, is this ok? It seemed like a formatted version of the date, though I notice that `dayjs` is used in `src/views/blog-post/post-title-header/post-title-header.astro` for `edited` & `published` dates.
2. I had to reverse the search when viewing earlier versions of a post (using `upToDateSlug` instead of slug). See [[postid].astro](https://github.com/ckhackshaw/playfulprogramming/blob/31022d0f8b76c6459c27823420ba6c1a97de8835/src/pages/%5B...locale%5D/posts/%5Bpostid%5D.astro#L44)
3. Did you want me to hide the current selected version from the version list? I left it because it seemed like bad UX to have the default date & version (i.e. when the page loads) on the button but not is the list

src/views/blog-post/article-revisions/article-revisions.module.scss
```diff
@@ -0,0 +1,173 @@
...
+	cursor: pointer;
+
+	&:hover {
+		background-color: var(--date-dropdown_state_hovered);
+		border-color: var(--date-dropdown_border-color_hovered);
+	}
+
+	&:focus {
+		outline: 4px solid #8c4585;
+		background-color: #fff;
```
fennifith:
```suggestion
		outline: var(--date-dropdown_focus-outline_width) solid var(--date-dropdown_focus-outline_color);
		background-color: var(--date-dropdown_background-color_focused)
```
Don't forget to swap these for the CSS variables so that it works in dark theme :)

src/views/blog-post/article-revisions/article-revisions.module.scss
```diff
@@ -0,0 +1,173 @@
...
+
+		.date {
+			color: #531350;
+		}
+
```
fennifith:
```suggestion
```

This has the same value as `--date-dropdown_foreground-color`, so I think it can be removed?

src/views/blog-post/article-revisions/article-revisions.module.scss
```diff
@@ -0,0 +1,173 @@
...
+	border-radius: var(--date-dropdown_corner-radius);
+	display: flex;
+	gap: var(--date-dropdown_gap);
+	align-items: center;
+	font-family: var(--pfp-font-family);
+	font-weight: var(--weight_semibold);
+	font-size: var(--button_regular_font-size);
+	line-height: var(--button_regular_line-height);
+	@include transition(background-color);
+	@include transition(border);
```
fennifith:
```suggestion
	@include transition(background-color border);
```
These styles override each other - can place them both in the same statement.

fennifith (commented):
> Relative and absolute positioning doesn't work here, however we can use [CSS anchor positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning), which I did. See browser support [here](https://developer.mozilla.org/en-US/docs/Web/CSS/position-anchor#browser_compatibility)

You're right that relative position won't work in the popover. Using anchor positioning would be a good workaround, but the firefox support is an issue (we should support [Baseline: Widely available](https://developer.mozilla.org/en-US/docs/Glossary/Baseline/Compatibility) where possible).

Could you use `CSS.supports()` to conditionally apply your previous JS positioning? Then the anchor positioning is still used on browsers that support it, but it can fall back to JS on browsers that don't.

> If you are happy with these constraints, let me know so that I can convert `ArticleRevisionDropdown` to an astro component

The other constraints seem good. I don't think it should be converted to astro, though! It can stay in preact :)

> I used the publishedMeta property for the dates, is this ok?

Yes, that's the intended use of publishedMeta! Not sure why the other locations aren't using it.

> Did you want me to hide the current selected version from the version list?

No, I think it's intended for the current version to be shown/pre-selected in the list. I'm not sure if the current item should be styled in the selected state, though - [this mockup](https://www.figma.com/design/Ul8KB1Sc6etCNTRCRSc27z/Playful-Programming---Website?node-id=8196-41037&t=nljkGO8aSIJ4nMpL-0) implies it should. (will confirm with Ed)

src/views/blog-post/post-title-header/post-title-header.astro
```diff
@@ -1,17 +1,19 @@
...
-import { PostInfo, PersonInfo } from "types/index";
+import { PostInfo, PersonInfo, PostVersion } from "types/index";
 import { Icon } from "astro-icon/components";
 import { Chip } from "components/index";
 import dayjs from "dayjs";
 import { buildSearchQuery } from "../../search/search";
 import { translate } from "utils/translations";
+import { ArticleRevisionDropdown } from "../article-revisions/article-revisions";
 
 interface PostTitleHeaderProps {
```
fennifith:
```suggestion
export interface Props {
```

Could rename this so that Astro's type checking works properly

src/views/blog-post/post-title-header/post-title-header.astro
```diff
@@ -1,17 +1,19 @@
...
 	post: PostInfo;
+	versions: PostVersion[];
 	authors: PersonInfo[];
 }
-const { post, authors } = Astro.props as PostTitleHeaderProps;
+const { post, versions, authors } = Astro.props as PostTitleHeaderProps;
```
fennifith:
```suggestion
const { post, versions, authors } = Astro.props;
```
With the above change, the `as` cast won't be needed

src/views/blog-post/post-title-header/post-title-header.astro
```diff
@@ -30,7 +32,17 @@ const originalLinkStr = post.originalLink && new URL(post.originalLink).host;
 			<div class={style.date}>
 				<div class={style.date__published}>
 					<Icon width={24} height={24} name="date" aria-hidden="true" />
-					<p class="text-style-button-regular">{publishedStr}</p>
+					{
+						versions.length ? (
```
fennifith:
```suggestion
						versions.length > 1 ? (
```
It seems like there's always be one version of each post, so the dropdown should only be shown if there's more than one.

fennifith (commented):
A few other nitpicks I missed:

fennifith:
On the selected item styling - I've realized it actually matches the [Popup Menu](https://www.figma.com/design/Ul8KB1Sc6etCNTRCRSc27z/Playful-Programming---Website?node-id=4579-43461&t=NlQMXM2abDvVsaAV-0) component. So the current page should have the "Default, Selected=true" background when it's unhovered, and use the other states in that list for hover/press/focus.

It might be better to reuse the styling from [select.module.scss](https://github.com/playfulprogramming/playfulprogramming/blob/main/src/components/select/select.module.scss#L61) and create a non-react-aria [`<Option>`](https://github.com/playfulprogramming/playfulprogramming/blob/main/src/components/select/select.tsx#L247) component with props for the anchor tag, href, and version label (to replace the checkmark). Then you don't have to reimplement its focus/hover states (this is probably why those aren't part of the date dropdown spec).

src/views/blog-post/article-revisions/article-revisions.module.scss
ckhackshaw: Done

src/views/blog-post/article-revisions/article-revisions.module.scss
ckhackshaw: Done

src/views/blog-post/post-title-header/post-title-header.astro
ckhackshaw: Done

src/views/blog-post/post-title-header/post-title-header.astro
ckhackshaw: Ditto

fennifith:
> The Popover option used [here](https://github.com/playfulprogramming/playfulprogramming/blob/17332957cd4439b62e83524908cff2df8147f145/src/components/select/select.tsx#L247) doesn't have any use for an anchor tag, so this messes with the styling.

You'll need to add to the styling from `select.module.scss`, but I don't think it needs to be duplicated.

I'd try something like 
```jsx
<li>
  <a class={`${styles.option} ${isSelected ? styles.selected : ""}`} href={...}>
    <span class={`text-style-button-regular ${styles.optionText}`}>...</span>
    <span class={`text-style-button-regular ${styles.optionPostfix}`}>...</span>
  </a>
</li>
```
and then add `.optionPostfix` to the option styles.

> Also, can you tell me what the View changelog text is supposed to do?

This should go to the article commits page (e.g. [this](https://github.com/playfulprogramming/playfulprogramming/commits/main/content/fennifith/posts/bevy-ecs-code-reusability/index.md)). It should be possible to piece that together from `post.path` (blog-post.astro has an example)

src/views/blog-post/post-title-header/post-title-header.astro
```diff
@@ -30,7 +32,17 @@ const originalLinkStr = post.originalLink && new URL(post.originalLink).host;
 			<div class={style.date}>
 				<div class={style.date__published}>
 					<Icon width={24} height={24} name="date" aria-hidden="true" />
-					<p class="text-style-button-regular">{publishedStr}</p>
+					{
+						versions.length > 1 ? (
+							<ArticleRevisionDropdown
+								client:only="preact"
```
fennifith: `client:only` should be avoided here. You can use `import.meta.env.SSR` in the component as a condition for any browser behavior, but it will create layout shift if it isn't pre-rendered :)

src/components/select/basic-option.tsx
```diff
@@ -0,0 +1,9 @@
+import { ComponentChildren } from "preact";
+
+interface OptionProps {
+	children: ComponentChildren;
+}
+
+export function Option({ children }: OptionProps) {
+	return <li>{children}</li>;
+}
```
fennifith: I think my suggestion here was misunderstood :sweat_smile: 

My idea was that the `<a>` tag and `<span>`s would live inside of `basic-option.tsx` so that they can avoid duplicating the styling from `option.module.scss`. Each option can then be wrapped in an outer `<li>` by `article-revisions.tsx` (since that is where the list is defined)

If you don't think that makes sense, I'd just remove this component since it isn't really doing anything.

src/views/blog-post/article-revisions/article-revisions.tsx
```diff
@@ -0,0 +1,132 @@
...
+							</span>
+							<span class={`text-style-button-regular ${style.version}`}>
+								{version}
+							</span>
+						</a>
+					</Option>
+				))}
+				<Option>
+					<hr class={style.divider} />
+				</Option>
```
fennifith: This `<hr>` might be an a11y issue since it would be read out as an option in the list, even though it's only a visual border.

I would style the final `<li>` and specify the border as a `border-top` style instead of using an element here.

src/views/blog-post/article-revisions/article-revisions.tsx
```diff
@@ -0,0 +1,132 @@
...
+				class={`${style.popover} ${supportsAnchors ? style.anchored : ""}`}
+				style={
+					supportsAnchors
+						? ""
+						: { left: `${popOverXY.x}px`, top: `${popOverXY.y}px` }
+				}
+			>
+				{versions.map(({ href, publishedMeta, version }, i) => (
+					<Option key={i}>
+						<a
```
fennifith, line 109: Another a11y issue remaining is that the options/links don't have focus styles when tabbed over.

The focus state is defined in [select.module.scss](https://github.com/playfulprogramming/playfulprogramming/blob/main/src/components/select/select.module.scss#L76) using the data-focus-visible attribute. However, since your component won't be javascript-controlled, you'll want to use `:focus-visible` instead.

If you do end up reusing those styles in `basic-option.tsx`, I'd add that as a `.option:not([data-focus-visible]):focus-visible` to avoid conflicting with the JS usage.

src/views/blog-post/article-revisions/article-revisions.tsx
```diff
@@ -0,0 +1,132 @@
...
+	const SPACING = 8;
+
+	const [supportsAnchors, setSupportsAnchors] = useState<boolean>(false);
+	const [popOverXY, setPopOverXY] = useState<PopOverLocation>({
+		x: 0,
+		y: 0,
+	});
+
+	useEffect(() => {
+		setSupportsAnchors(CSS.supports("top: anchor(bottom)"));
```
fennifith: `supportsAnchors` doesn't need to be in a useState, as it will never changed once declared. I'd add a `const supportsAnchors = import.meta.env.SSR || CSS.supports("top: anchor(bottom)");` outside of the component and remove the `useState`. This will help avoid a re-render when the page loads.

fennifith (commented):
This is good progress! Left a few more comments :)

crutchcorn (approved):

Issue #1233, opened 2024-12-05
Only Load Relevant <script> tags based on their usage in MD


crutchcorn:
The idea is that if `tabs` isn't used, it shouldn't have `tabs` as a script that's loaded on the blog post page.

Should be conditional based on its usage

crutchcorn:
Work has begun in this branch: https://github.com/playfulprogramming/playfulprogramming/tree/conditional-script-usage

PR #1301, opened 2025-02-13
[WIP] [Ready for Review] Conditional script usage

crutchcorn:
This PR fixes the issue outlined in #1233 where, say, `tabs`, loads the runtime JS for the tabs component, even if the component is not used in the blog post itself.

As part of this, I also restructured the architecture and DX API for creating a component. Now, there's a consistent util to make a component:

```typescript
export const tabs = createComponent<Record<string, never>>()
	.withBuildTime({
                // Only returns props to be passed to `componentFSPath`'s default function
		transform: transformTabs,
	})
	.withRuntime({
		componentFSPath: resolve(__dirname, "tabs.tsx"),
                // Optional
		setup: enableTab,
                // Optional
                takedown: disableTab
	});
```

This allows us some flexibility with:

- How we spin up and take down a runtime script
    - "takedown" is not used in the blog site, but will be used in the CMS editor to display and make interactive the components in the rich-text editor
- Which runtime we use for JSX/TSX (hastscript for blog, Preact for in-editor CMS)


taylor-shift (approved):

src/views/blog-post/blog-post.astro
```diff
@@ -46,20 +48,32 @@ const {
...
+	usedComponents,
 } = Astro.props;
 
 const GHLink = `https://github.com/${siteMetadata.repoPath}/tree/main/content/${post.path}/index.md`;
+
+const scriptServerPaths = Object.keys(usedComponents).reduce((prev, key) => {
+	if (!usedComponents[key]) return prev;
+	prev.push(getComponentScriptPath(key).serverRelativePath);
+	return prev;
+}, [] as string[]);
```
20jasper, line 60: nit: consider passing a type argument to avoid `as` casting

```diff
- const scriptServerPaths = Object.keys(usedComponents).reduce((prev, key) => {
+ const scriptServerPaths = Object.keys(usedComponents).reduce<string[]>((prev, key) => {
	  if (!usedComponents[key]) return prev;
	  prev.push(getComponentScriptPath(key).serverRelativePath);
	  return prev;
+ }, []);
- }, [] as string[]);
```

src/utils/markdown/components/rehype-transform.ts
```diff
@@ -102,31 +147,67 @@ export const rehypeTransformComponents: Plugin<
...
+
+					// Recursively transform the children
+					// This allows for nested components like ebook only content in tabs
+					await Promise.all(
+						replacementArray.map(async (replacement) => {
+							if (!isNodeParent(replacement)) return;
+							await (
+								rehypeTransformComponents as (
+									props: RehypeComponentsProps,
+								) => (tree: Root, vfile: VFile) => void
```
20jasper, line 196: Should `rehypeTransformComponents` be cast like so? Consider one of the following 

If the final function returns a Promise, then it can be awaited
```diff
- (props: RehypeComponentsProps) => (tree: Root, vfile: VFile) => void;
+ (props: RehypeComponentsProps) => (tree: Root, vfile: VFile) => Promise<void>;
```
Otherwise, the await will have no effect, which will effectively return `Promise<void>`
```diff
-           await (
+           return (
								rehypeTransformComponents as (
									props: RehypeComponentsProps,
				 				) => (tree: Root, vfile: VFile) => void
			 				)({ components })(replacement as Root, vfile);
```

src/views/blog-post/blog-post.astro
crutchcorn: I don't prefer this syntax as it makes things harder to read IMO and breaks implicit return typing of the function itself.

fennifith:
Replaced by #1350 :)

PR #1306, opened 2025-02-24
Mobile Table of Contents 

josephmfaulkner:
This item has now been fully implemented. Just some refactoring is left to do.


20jasper:
@josephmfaulkner https://github.com/playfulprogramming/playfulprogramming/pull/1295

For reference

20jasper:
Notes for what we talked about
- use an intersection observer on the `h1` to reset to table of contents
- use a selector like the following to `main:not(:has(> .mobile-menu)) ~ header` to remove bottom border when mobile menu exists
- use similar functionality to https://github.com/playfulprogramming/playfulprogramming/pull/1295 to scroll to the middle of the page for parity with existing table of contents. Consider making a more generic function that takes in a parent selector instead of just a `.post`

src/views/base/navigation/header.module.scss
```diff
@@ -22,15 +22,22 @@ header {
 }
 
 .headerBorder {
+	background-color: var(--background_primary);
```
josephmfaulkner, line 25: I moved this since in some cases, the article's contents was overlapping with the header.

src/views/blog-post/table-of-contents/heading-intersection-observer-script.astro
```diff
@@ -24,7 +24,7 @@ const { headingsToDisplaySlugs } = Astro.props as {
 		const tocListContainer = tocListRef.parentElement!.parentElement!;
 
 		const linkRefs: LinkRef[] = [
-			...(document.querySelectorAll(
+			...(tocListRef.querySelectorAll(
```
josephmfaulkner, line 27: This makes sure the linkRefs is only populated with items inside the list component, rather than the whole document.

crutchcorn (approved):
This code looks great to me! Thank you so much @josephmfaulkner - this is a great value add.

Let me deploy it to a preview branch, run tests, and we can merge if all looks good!

crutchcorn (changes requested):
This is **such** a fun addition!

Three things we need to fix before going live:

1) Tapping anywhere on the mobile TOC should open it, not just the button
2) We need to edgecase where a blog post has `0` headings like this one:

https://playfulprogramming-git-fork-josephmfaulkner-mob-2587e3-oceanbit.vercel.app/posts/a-forgotten-idiom-revisited-pass-key

image

Otherwise we get a weird look. I vote we remove the mobile header entirely for blog posts like this

3) The header bottom line seems to be broken with this PR:

image

josephmfaulkner:
Hey @crutchcorn, the first two changes you've requested make sense to me. I'm afraid I'm unable to reproduce the third issue you're having. Do you have specific reproduction steps for me? Which browser are you using? I've tried both Chrome and Firefox. 
https://playfulprogramming-git-fork-josephmfaulkner-mob-2587e3-oceanbit.vercel.app/

crutchcorn:
Interesting. I can still reproduce that issue from my end with that link - I'm on Chrome `134.0.6998.118`

josephmfaulkner:
Okay, I've got two commits in for the first two issues.

src/views/blog-post/table-of-contents/mobile/table-of-contents-mobile.module.scss
```diff
@@ -0,0 +1,294 @@
...
+		.tocTitle {
+			button {
+				.mobileTocOpenIcon { display: inline; }
+				.mobileTocCloseIcon { display: none;  }
+			}
+		}
+
+		.tableList {
+			height: 0px;
+			padding: 0;
```
fennifith, line 120: This `padding: 0` means that the content abruptly shifts to zero padding when closing the menu. The scrollbar also causes layout shift when it appears/disappears during this animation (depends on the browser; yours may or may not have any effect on the layout). I think these are both limitations of animating by `height`

What I'd suggest here is actually animating between `clip-path: inset(0)` (fully visible) and `clip-path: inset(0 0 100% 0)` (fully hidden) instead - this way no layout changes occur during the animation. You'll also need to set `position: absolute;` on the `.tableList` for this to work.

src/views/blog-post/blog-post-layout/blog-post-layout.module.scss
```diff
@@ -87,3 +87,18 @@
 		opacity: 1;
 	}
 }
+
+.mobileMenu {
```
fennifith, line 91: I've noticed that, while the menu is expanded, it is still possible to scroll the page behind it by swiping on mobile - and this also changes the selected heading while the ToC is open (even though you can't see the content behind it).

Applying `overflow: hidden` to the `<body>` should prevent this - and that _won't_ cause layout shift in this case because we have `scrollbar-gutter: stable` in our CSS :)

(the search page / filter modals previously ran into this same issue)

fennifith (commented):
This looks great so far! Commented on few things I found poking around.

src/utils/sticky-observer-script.ts
```diff
@@ -7,15 +7,18 @@
...
+		.querySelectorAll<HTMLElement>("[data-sticky-observer]")
+		.forEach((el: HTMLElement) => {
+			new IntersectionObserver(
+				([e]) => {
+					(e.target as HTMLElement).dataset.sticky =
+						e.intersectionRatio < 1 ? "pinned" : "";
+				},
+				{
+					threshold: [1],
+					rootMargin: 0 - parseInt(getComputedStyle(el).top) + "px",
```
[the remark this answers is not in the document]
josephmfaulkner: This change has been superseded by this PR: https://github.com/playfulprogramming/playfulprogramming/pull/1314

src/views/blog-post/table-of-contents/mobile/table-of-contents-mobile.module.scss
josephmfaulkner: https://github.com/josephmfaulkner/playfulprogramming/commit/76faae41d6cfad4c7348dc92b05b505750b1bf80

src/views/blog-post/blog-post-layout/blog-post-layout.module.scss
josephmfaulkner: https://github.com/josephmfaulkner/playfulprogramming/commit/7befaca0334e58af2f9cfe39acc732ad02308a2d

fennifith (approved):
LGTM!

PR #1336, opened 2025-05-21
[WIP] Upgrade chapterlist feature

### LINKED ISSUE #1175: [Feature]: Upgrade & fix the chapterList feature
### Description

The current `chapterList` feature accepts the following props:

```js
chapterList: [
  {
    order: '',
    title: "",
    description: "",
  },
]
```

However, when trying to use this for a real-world scenario - embedding content from the website, outside of a collection, into a collection - the feature did not help me, and in fact, didn't work as expected.

### Main issues
- **The `order` prop is meaningless.** It is not taken into account when ordering content and instead, all posts added with `chapterList` are appended at the bottom, making the feature essentially useless for anything other than the only current usecase: A coming soon `?` as the value.
- **No `url` prop.** Essentially, every entry is just a string that does not point to anything, and the lack of a `url` prop makes it impossible to turn it into a link either.

### Feature requests
- Allow the use of a single `url` prop to link to articles outside of collections.

The syntax for this would look as such: 

```js
chapterList: [
  {
    url: '/posts/understanding-the-dom'
    // Extracts the title and snippet automatically
  },
]
```

- **When given a `url`, all other props act as overrides.**

This is a real need I ran into when attempting to link an article from outside a collection - something I think should be a native feature, as collections should, conceptually, be a global grouping feature as opposed to the `author > collection` hierarchy we currently have.

The idea is that, when writing an article, you may want to focus on SEO, and that drives the way you title and the descriptions you write in order to maximize that. But when attaching an article to a collection, the idea is to guide the reader, and provide context within the collection for the next chapter.

With this in mind, it would be useful if the `title` and `description` acted as overrides when a `url` is given. This would allow for contextual guidance / consistency with other chapters while maintaining the existing SEO.

### Describe the solution you'd like

_No response_

### Describe alternatives you've considered

_No response_

ckhackshaw:
Fixes #1175

This is the proposed solution for the issue, however there are a few things I should mention as I am not sure if these are part of the solution.

I have added two types to handle post internal & external posts as per my messages in our pfp-dev discord chat, however the original ChapterList type is still in use in the FFGF TOC `(src/views/collections/framework-field-guide-fundamentals/components/collection-table-of-contents-fundamentals.astro)` file located in the views folder.

On the posts page, the `SeriesToC` component uses the `postSeries` prop which is of type `PostInfo[]`. This is different to the chapter list but appears to render the same thing. Here is the file I am talking about here `src/views/blog-post/series/series-toc.astro`. Same issue with the `ArticleNav` component.


src/pages/[...locale]/collections/[slug].astro
```diff
@@ -85,14 +92,32 @@ const collectionHtml = await getMarkdownHtml(collection);
...
+								const postData = api.getPostBySlug(post.post, locale);
+								title = postData?.title || "No title";
+								description = postData?.description || "No description";
+								link = post?.post ? `/posts/${post.post}` : "#";
+							} else {
+								title = post.title;
+								description = post.description;
+								link = post.url;
+								externalLink = true;
+							}
```
Comment (fennifith) [inline src/pages/[...locale]/collections/[slug].astro:111]: More of a formatting nit - can we move this logic (+ what I've mentioned above) to the top of this file to avoid the nesting?

src/pages/[...locale]/collections/[slug].astro
```diff
@@ -85,14 +92,32 @@ const collectionHtml = await getMarkdownHtml(collection);
 						})
 					}
```
Comment (fennifith) [inline src/pages/[...locale]/collections/[slug].astro:93]: The new `chapterList` behavior should ideally replace the above list of actual posts when it is present. In other words, `collectionPosts` should only be used when `chapterList` is not defined.

fennifith (commented):
This looks great so far! Left a few comments, and answering questions here:

- For the FFGF, I don't think that should be changed as part of this - probably a better solution is to copy the previous `ChapterList` type into `collection-table-of-contents-fundamentals.astro` so that it can be used independently.
- The `SeriesToC` list on the posts page makes some different assumptions - IMO it's better to only list "real posts" there so that the UX is clear (and matches the next/previous buttons on the post layout)

src/pages/[...locale]/collections/[slug].astro
Comment (ckhackshaw) [reply src/pages/[...locale]/collections/[slug].astro:111]: Done

src/pages/[...locale]/collections/[slug].astro
```diff
@@ -45,6 +54,51 @@ const authors = collection.authors.map(
...
+
+const isLocalPost = (
+	chapter: LocalPost | ExternalPost,
+): chapter is LocalPost => {
+	return (chapter as LocalPost).post !== undefined;
+};
+
+let collectionData: Array<CollectionChapterData> = [];
+
+if (collection?.chapterList?.length) {
```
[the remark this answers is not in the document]
Comment (ckhackshaw) [reply src/pages/[...locale]/collections/[slug].astro:66]: Made these changes

Note that when `chapterList` is empty and the `collectionPosts` are used, the numbering can be inconsistent since `PostInfo.order` is optional.

I ended using the loop index if `order` is undefined.

fennifith (approved):

PR #1342, opened 2025-05-26
Improve search page a11y and seo

### LINKED ISSUE #856: Move search `h1` out of Preact island
Right now, Preact is loading for the search page client-only, which makes sense for the search page.

However, this has led to an SEO/A11Y issue when JS is off:

image

We should also probably have [a `noscript` option to show a warning when JS is disabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/noscript)

ckhackshaw:
Fixes #856

This PR moves the `main` & `h1` elements outside the `search-page.tsx` component to the `search.astro` component to improve a11y and SEO.

An effect is now used to toggle the change when the sidebar is shown

### DIFF
--- a/src/pages/search.astro
+++ b/src/pages/search.astro
@@ -1,4 +1,5 @@
 ---
+import style from "../views/search/search-page.module.scss";
 import SearchPage from "../views/search/search-page";
 import Document from "../layouts/document.astro";
 import SEO from "../components/seo/seo.astro";
@@ -13,6 +14,11 @@ import { siteMetadata } from "constants/site-config";
 	/>
 	<main>
 		<h1 class="visually-hidden">Search</h1>
+		<noscript>
+			<h3 class={style.centerHeader}>
+				JavaScript needs to be enabled to use this page
+			</h3>
+		</noscript>
 		<SearchPage client:only="preact" siteTitle={siteMetadata.title} />
 	</main>
 </Document>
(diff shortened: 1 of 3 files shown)


src/views/search/search-page.tsx
```diff
@@ -322,11 +333,7 @@ export function SearchPageBase({ siteTitle }: RootSearchPageProps) {
 	const numberOfPosts = showArticles ? data.totalPosts : 0;
 
 	return (
-		<main
-			className={style.fullPageContainer}
-			data-hide-sidebar={!query.searchQuery}
-		>
-			<h1 className={"visually-hidden"}>Search</h1>
+		<>
```
fennifith, line 336: Could the `useEffect` be avoided by making this a `<div>` to hold the data-hide-sidebar attribute instead? (and removing the styles from the`<main>` element in search.astro)

```suggestion
		<div
			className={style.fullPageContainer}
			data-hide-sidebar={!query.searchQuery}
		>
```

ckhackshaw: Done.

src/pages/search.astro
```diff
@@ -11,5 +12,13 @@ import { siteMetadata } from "constants/site-config";
 		description="Find your favorite articles and collections on Playful Programming easier than ever!"
 		slot="head"
 	/>
-	<SearchPage client:only="preact" siteTitle={siteMetadata.title} />
+	<main>
+		<h1 class="visually-hidden">Search</h1>
+		<noscript>
+			<h3 class={style.centerHeader}>
```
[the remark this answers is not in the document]
ckhackshaw: Done

fennifith (approved):

PR #1350, opened 2025-06-09
Separate interactive components from the markdown AST

### LINKED ISSUE #1233: Only Load Relevant <script> tags based on their usage in MD

fennifith:
Fixes #1233, but differently™

This PR changes the unified processor to return a "component tree" rather than a plain HTML string. This tree contains portions of raw HTML as well as props that would normally be passed to interactive components (tabs, file lists, code embeds, link previews, etc).

The component tree is then passed to an Astro component, which is used recursively to render the raw HTML and components.

This results in any imports (styles & scripts) being conditional on the component appearing in the post. It also means that scoped CSS and client hydration are supported.

In the future, this component tree could be cached with the post content, and any SSR-dependent behavior could happen in the Astro components.

## Changes to note:

- Components can now _only_ be nested at the root level of other components, or within specific elements that are transformed into components (`<details>`)

  This means that iframes can no longer be placed within a list, as was done in `example/index.md`. When this happens, the transformer will log & throw an error.

- EPub generation is moved from a separate build script to a `pages/[slug].epub.ts` Astro route
  (this means it isn't parallelized, but takes advantage of caching & will massively improve the build time with #1335)

- Conditional `tabs-small` behavior (keeping tabs at a fixed height) is removed

### TODO:

- [x] Migrate tabs to preact
- [x] Migrate in-content-ad to preact
- [x] Migrate link-preview to preact
- [x] Migrate filetree to preact
- [x] Migrate IFramePlaceholder to preact
- [x] Migrate hints to preact
- [x] Validate epub changes

### DIFF
--- a/src/pages/[...locale]/collections/[slug].astro
+++ b/src/pages/[...locale]/collections/[slug].astro
@@ -1,7 +1,6 @@
 ---
 import SEO from "src/components/seo/seo.astro";
 import Document from "src/layouts/document.astro";
-import { MarkdownInstance } from "astro";
 import * as api from "utils/api";
 import CollectionPage from "src/views/collections/collection-page.astro";
 import CollectionChapter from "src/views/collections/collection-chapter.astro";
@@ -26,20 +25,6 @@ export async function getStaticPaths() {
 			};
 		});
 }
-
-interface CollectionProps {
-	Content: MarkdownInstance<never>["Content"];
-}
-
-interface CollectionChapterData {
-	num?: number;
-	title: string;
-	description: string;
-	href: string;
-	externalLink: boolean;
-}
-
-const { Content } = Astro.props as CollectionProps;
 
 const params = Astro.params as {
 	locale?: Languages;


src/utils/markdown/components/components.ts
```diff
@@ -0,0 +1,88 @@
...
+export const components = {
+	FileList,
+	InContentAd,
+	LinkPreview,
+	Tabs,
+	Tooltip,
+	IframePlaceholder,
+	Hint,
+} as const;
+
```
20jasper, line 76: Personal opinion that I will not block on

Consider defaulting children inline in the function parameter

Consider pulling `Parameters<(typeof components)[Key]>[0]` into a defaulted type or into a type alias

```diff
- export function createComponent<Key extends keyof typeof components>(
+ export function createComponent<Key extends keyof typeof components, NodeProps = Parameters<(typeof components)[Key]>[0]>(
  	key: Key,
-  	props: Omit<Parameters<(typeof components)[Key]>[0], "children">,
+  	props: Omit<NodeProps, "children">,
   	children?: ComponentNode["children"],
- ): ComponentNode<Parameters<(typeof components)[Key]>[0]> {
+ ): ComponentNode<NodeProps> {
```

src/utils/smooth-scroll-for-anchors-to-current-page.ts
```diff
@@ -4,7 +4,6 @@ export const enableSmoothScrollForAnchorsToCurrentPage = () => {
 		window.matchMedia("(prefers-reduced-motion: reduce)").matches;
 
 	const els = Array.from(document.querySelectorAll(`.post-body a[href^="#"]`));
-	console.log(els);
```
20jasper, line 7: Rest In Peace console log. You have served us well :saluting_face:

src/utils/markdown/iframes/rehype-transform.ts
```diff
@@ -204,32 +220,42 @@ export const rehypeUnicornIFrameClickToRun: Plugin<
...
+						pageIcon: info?.iconFile,
+						iframeAttrs: Object.fromEntries(
+							Object.entries(propsToPreserve).map(([key, value]) => [
+								key,
+								// Handle array props per hast spec:
+								// @see https://github.com/syntax-tree/hast#propertyvalue
+								Array.isArray(value) ? value.join(" ") : String(value),
+							]),
+						),
+					}) as never,
```
20jasper: Is this `as never` necessary? Looks like it's the right type to me

src/components/file-list/file-list.tsx
```diff
@@ -0,0 +1,104 @@
...
+					</span>
+					{item.name}
+				</span>
+				{item.commentHtml && item.commentHtml.length ? (
+					<span class={`${style.comment} text-style-body-small`} dangerouslySetInnerHTML={{ __html: item.commentHtml }} />
+				) : null}
+			</summary>
+			{FileListList({ items: item.items })}
+		</details>
+	) as never;
```
20jasper: Is this `as never` necessary? Everywhere it's used seems not to complain if it has the correct type

src/pages/[...locale]/collections/[slug].astro
```diff
@@ -118,7 +118,7 @@ if (collection?.chapterList?.length) {
 	<main>
```
Comment (20jasper) [inline src/pages/[...locale]/collections/[slug].astro:103]: No problem with this, just wanted to call out I noticed lines 30 to 42 are unused in this file. Maybe worth handling in another PR?

src/utils/markdown/components/components.ts
fennifith: I had attempted this exact way of writing it initially! For reasons I can't explain, it completely breaks the type checking and causes the function to just accept any object as `props` regardless of whether it matches the component properties or not.

I'll take your suggestion on defaulting the `children` param though :)

src/utils/markdown/iframes/rehype-transform.ts
fennifith: Yeah, there were a lot of `never` casts floating around from previous (badly typed) implementations. Nice catch!

src/pages/[...locale]/collections/[slug].astro
Comment (fennifith) [reply src/pages/[...locale]/collections/[slug].astro:103]: Fixed!

src/utils/markdown/components/components.ts
20jasper: Hmm, maybe an `extends` is in order?

PR #1366, opened 2025-07-06
Pin astro to 5.7 to 5.8

20jasper:
Pin to astro 5.7 since 5.8 enforces a node version not supported by Vercel (latest version of node 18)

https://astro.build/blog/astro-580/#updated-minimum-nodejs-version

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
 		"@typescript-eslint/eslint-plugin": "^8.21.0",
 		"@typescript-eslint/parser": "^8.21.0",
 		"@vercel/node": "^5.0.2",
-		"astro": "^5.7.14",
+		"astro": "5.7.14",
 		"astro-eslint-parser": "^1.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -69,7 +69,7 @@
 		"@typescript-eslint/eslint-plugin": "^8.21.0",
 		"@typescript-eslint/parser": "^8.21.0",
 		"@vercel/node": "^5.0.2",
-		"astro": "^5.1.8",
+		"astro": "^5.7.14",
```
fennifith:
```suggestion
		"astro": "5.7.14",
```
We'd want to remove the "^" in order to fully pin the version, right? Otherwise it could still get updated to 5.8 in the lockfile

20jasper: I have no clue why I left it as a caret constraint--you're right either we'd want a tilde constraint or pin an exact version. I personally like tilde constraint better but an exact version is also fine!

PR #1382, opened 2025-08-17
Improve fullstack dev environment

mikecebul:
## Summary
If testing out builds locally is required then we need to run against a local development instance of Hoof. To accomplish this I 

- created a local.env file 
- added a REPO_PATH var to test out Stackblitz iframes in dev
- created dev:local and build:local scripts using those env vars
- added a fallback in production to use local files for pictures if there is no cloudinaryCloudName
- added documentation on setting this up

This allows to fully test out the Astro frontend. 

## Issues
Building the first time took me 42 minutes on my M1 Macbook. Once cached it build in one minute. We should look into creating a seed script in the future.

### DIFF
--- a/src/environments/local.env
+++ b/src/environments/local.env
@@ -3,4 +3,4 @@ NODE_OPTIONS="--max-old-space-size=8192"
 HOOF_URL=http://localhost:3000
 
 # For functional stackblitz iframes add your Github Repo name for the path
-REPO_PATH=yourname/playfulprogramming
+REPO_PATH=mikecebul/playfulprogramming
--- a/src/environments/local.env
+++ b/src/environments/local.env
@@ -1,0 +1,6 @@
+BUILD_ENV="development"
+NODE_OPTIONS="--max-old-space-size=8192"
+HOOF_URL=http://localhost:3000
+
+# For functional stackblitz iframes add your Github Repo name for the path
+REPO_PATH=yourname/playfulprogramming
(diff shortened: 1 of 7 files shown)


src/environments/local.env
```diff
@@ -0,0 +1,6 @@
+BUILD_ENV="development"
+NODE_OPTIONS="--max-old-space-size=8192"
+HOOF_URL=http://localhost:3000
+
+# For functional stackblitz iframes add your Github Repo name for the path
+REPO_PATH=mikecebul/playfulprogramming
```
fennifith:
```suggestion
# For functional stackblitz iframes add your Github Repo name for the path
REPO_PATH=playfulprogramming/playfulprogramming
```

This should probably be the main repo path in git. Might want to add an untracked env file so that local changes don't get committed here (or make a root `.env` for that purpose)

Also, FYI the stackblitz iframes themselves won't exist for much longer - #1377 :)
(this is still useful, though!)

mikecebul: Thanks for the feedback! Ok, I'm adding a root .env and updating docs.

mikecebul: I ran out of time for today, i'll get back to it later.

mikecebul: I see your pr #1377 which makes the extra logic of making them work in dev unnecessary. I reverted a commit of me messing with env vars and completely removed the use of REPO_PATH env in anticipation of the the new web component.

Issue #1379, opened 2025-08-12
[Website Bug]: collections have right margin


20jasper:
### Description

The page loads in correctly, then resizes itself onload to have a bit of right margin

Iphone 12 mini, iOS 17.7, Safari browser

<img width="1125" height="2436" alt="A stripe of empty is to the right on the playful programming site" src="https://github.com/user-attachments/assets/a6ad4a3b-4e67-4241-8050-9b39c01b0629" />

### Minimal Reproduction

Go to https://playfulprogramming.com/collections/framework-field-guide

PR #1399, opened 2025-09-10
Fix layout width issues

ckhackshaw:
Fixes the issue where on smaller mobile devices, there would be a space to the right of the page that looks like a margin. See #1379

I was not able to replicate the bug on my device because i have a larger screen than the iPhone 12 @20jasper found the bug on, but I did using the inspector. However I had to make the screen a lot smaller (>320 px) to get the same effect.

@20jasper, can you confirm that the issue is fixed on the device? The issue here is that some elements (@crutchcorn metrics & the section on the 'the adventure continues') were wider than the available width.


ckhackshaw:
> Ping me again when it's deployed to vercel and I can check!

It's ready

fennifith (approved):

PR #1418, opened 2025-09-30
Modify details summary in CSS specificity post

superhighfives:
Hey there,

I believe the specificity of this example is wrong. Earlier in the post you use `.class.class.class` as `0-3-0`, so this should be the same.

I also tested this against the https://specificity.keegan.st/

image


ckhackshaw:
@crutchcorn yes this was a mistake on my part sorry.. You can merge

Good catch @superhighfives, thanks a million

PR #1463, opened 2025-12-31
Closing submissions

Aaday321:
Added a `classIsFull` state

This is controlled by a hardcoded boolean:
```javascript
let classIsFull = true
```
When `true`:
Screenshot 2025-12-31 at 2 47 38 PM

When `false`:
Screenshot 2025-12-31 at 2 45 27 PM


Aaday321:
View this PR first, I think this approach might be better:
https://github.com/playfulprogramming/playfulprogramming/pull/1464

PR #1467, opened 2026-01-06
Add eslint preact

20jasper:
Missing dependencies in dependency arrays have caused flakey test issues (as seen in #1446 and #1466). This will catch these automatically, as well as other common issues like missing keys

- lints tsx files
- adds preact eslint rules


src/utils/smooth-scroll-for-anchors-to-current-page.ts
```diff
@@ -13,7 +13,8 @@ export const enableSmoothScrollForAnchorsToCurrentPage = () => {
 			const block =
 				((e.target as HTMLAnchorElement | null)?.getAttribute(
 					"data-scroll-block",
-				) as ScrollLogicalPosition) ?? "center";
+					// eslint-disable-next-line no-undef
```
20jasper, line 16: Not sure why eslint doesn't know about ScrollLogicalPosition

src/views/blog-post/table-of-contents/mobile/mobile-heading-intersection-observer-script.astro
```diff
@@ -9,6 +9,11 @@ const { headingsToDisplaySlugs } = Astro.props as {
 </script>
 
 <script>
+	declare global {
```
20jasper, line 12: There's probably a better way to do this than attaching to the window, like adding a json script maybe?

src/views/search/search-page.ui.spec.tsx
```diff
@@ -1033,49 +1030,59 @@ describe("Search page", () => {
 			authors: { [MockPerson.id]: 1 },
 		}));
 
-		var { getByTestId, getByText } = render(
-			<SearchPage mockClients={clients} />,
-		);
+		{
```
20jasper, line 1033: var begone!

PR #1497, opened 2026-01-14
dockerize e2e tests

20jasper:
Dockerizes e2e tests to allow non debian users to update snapshots locally and minimize differences between debian versions. The dev server is also dockerized to reduce local differences in astro config, for example if the dev toolbar is showing

## performance

Dockerizing only adds around 30 seconds to the e2e test time, leaving it at around 3 minutes. In the future we can cache builds if we so choose.  Git LFS is used for the snapshots since each is over 1MB. Only relevant GitHub Actions will pull these images

The server files and content are setup as a docker volume for convenience and speed of local development. Production builds take far too long for a test run, and a full container hurts the iteration speed

## Usage

Usage instructions from the included documentation

> # E2E tests Quick start
> 
> These tests run against the web service started by Docker Compose.
> 
> - Run tests: `pnpm test`
> - Update snapshots: `pnpm test:update-snapshots`
> 
> If you'd like to pass extra args to playwright, set the `PLAYWRIGHT_ARGS` environment variable
> 
> ```
> PLAYWRIGHT_ARGS="--update-snapshots" pnpm test
> ```

## alternatives

Also handles the following
- **chore: don't format mock service worker files**

## TODO
- [x] make sure Astro dev toolbar is not in screenshots
- [x] document git LFS usage
- validate cross platform usage
  - [x] Linux
  - [x] Mac
  - [ ] WIndows

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -6,8 +6,7 @@
   "type": "module",
   "scripts": {
     "test": "docker compose -f ./docker-compose.e2e.yml up --abort-on-container-exit --exit-code-from e2e --build",
-    "test:update-snapshots": "cross-env PLAYWRIGHT_ARGS=--update-snapshots pnpm test",
-    "test:update": "pnpm test:update-snapshots"
+    "test:update-snapshots": "cross-env PLAYWRIGHT_ARGS=--update-snapshots pnpm test"
   },
   "devDependencies": {
     "@playwright/test": "^1.57.0",
--- a/package.json
+++ b/package.json
@@ -41,6 +41,9 @@
 		"test:unit": "vitest",
 		"test:unit:node": "vitest --project node",
 		"test:unit:ui": "vitest --project ui",
+		"test:e2e": "pnpm --filter e2e test",
+		"test:e2e:update-snapshots": "pnpm --filter e2e test:update-snapshots",
+		"test:e2e:update": "pnpm --filter e2e test:update-snapshots",
 		"test:debug": "set DEBUG_PRINT_LIMIT=9999999999 && pnpm vitest",
 		"test:stress": "./scripts/run-tests-concurrently.sh --run",
 		"prepare": "husky"


e2e/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM mcr.microsoft.com/playwright:v1.57.0-jammy
+
+WORKDIR /work
+
+RUN corepack enable && corepack prepare pnpm@9.4.0 --activate
```
fennifith:
```suggestion
# Prepare pnpm according to the root package.json
COPY package.json .
RUN corepack enable && corepack install
```
This will prevent the Dockerfile from becoming out of date with the version specified in package.json

e2e/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM mcr.microsoft.com/playwright:v1.57.0-jammy
+
+WORKDIR /work
+
+ENV COREPACK_HOME=/usr/local/share/corepack
+
+RUN corepack enable && \
+  corepack prepare pnpm@9.4.0 --activate && \
+  chmod -R 755 $COREPACK_HOME || chmod -R 755 /root/.cache/node/corepack 2>/dev/null || true
```
fennifith:
```suggestion
COPY package.json .
RUN corepack enable && \
  corepack install && \
  chmod -R 755 $COREPACK_HOME || chmod -R 755 /root/.cache/node/corepack 2>/dev/null || true
```
I'd prefer `corepack install` here, as it uses the `packageManager` config from package.json - and prevents them from getting out of sync :)

e2e/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN pnpm install --filter e2e --force
+
+COPY e2e/ ./e2e/
+
+WORKDIR /work/e2e
+
+RUN mkdir -p .playwright/test-results .playwright/report \
+  && chmod -R a+rwX .playwright
+
+CMD ["pnpm", "test"]
```
fennifith:
```suggestion
```

Nit: Can we remove this `CMD`? Doesn't seem like it's being used

e2e/tests/posts-example.spec.ts-snapshots/posts-example-renders-dark-mode-1-chromium-linux.png
crutchcorn, line 1: Image

I noticed that it still has the Astro toolbar that we probably want to remove

e2e/README.md
```diff
@@ -0,0 +1,13 @@
+# E2E tests Quick start
+
```
crutchcorn, line 4: Can we add docs about LFS usage here? It's not obvious since the rest of the repo doesn't use it

crutchcorn (commented):
This doesn't yet work on macOS, but should be a minor fix:

```
web-1  | Error: Cannot find module @rollup/rollup-linux-arm64-gnu. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
```

Image

e2e/Dockerfile
20jasper: Makes sense!

e2e/Dockerfile
20jasper: Sure

e2e/tests/posts-example.spec.ts-snapshots/posts-example-renders-dark-mode-1-chromium-linux.png
20jasper: Good catch! The CI=true got lost somewhere. I added it back

e2e/README.md
20jasper: Good callout

e2e/tests/posts-example.spec.ts-snapshots/posts-example-renders-dark-mode-1-chromium-linux.png
20jasper: I tried a few things, and for whatever reason, this toolbar pops up on dark mode only
1. setting devtoolbar to false in the astro config
2. disabling globally with the astro CLI

Any ideas?

e2e/Dockerfile
20jasper: Done

e2e/Dockerfile
20jasper: Done

e2e/README.md
20jasper: Done

astro.config.ts
```diff
@@ -68,8 +71,4 @@ export default defineConfig({
 		},
 	},
 	markdown: {} as AstroUserConfig["markdown"] as never,
-	devToolbar: {
```
20jasper, line 71: Moved to dockerfile

public/mockServiceWorker.js
```diff
@@ -7,163 +7,163 @@
  * - Please do NOT modify this file.
  */
 
-const PACKAGE_VERSION = "2.12.7";
```
20jasper, line 10: This is autogenerated, so removed from formatting rules

e2e/playwright.config.ts
```diff
@@ -69,14 +79,4 @@ export default defineConfig({
 		//   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
 		// },
 	],
-
```
20jasper, line 72: Server lifecycle handled by docker compose now

PR #1514, opened 2026-01-19
Mock external services in e2e tests

### LINKED ISSUE #1511: Enable mocking e2e test data
Currently, the e2e screenshot tests depend on the dev server rendering the `/posts/example` post. This depends on a number of external APIs that can make it inconsistent between CI and local development.

We should create a config or dev mode that substitutes these services with mock data instead of the real backends. This includes:
- /tasks/url-metadata on Hoof
- /tasks/post-images on Hoof
- the GitHub graphql endpoint
- OEmbed providers / video embed data

The end result should be that `/posts/example` can be used in e2e tests without needing external internet access at all.

fennifith:
Fixes #1511


playwright.config.ts
```diff
@@ -73,7 +73,8 @@ export default defineConfig({
 	/* Run your local dev server before starting the tests */
 	webServer: {
 		/* CI=1 tells astro.config.ts to turn off the dev toolbar */
-		command: "CI=1 pnpm dev --port 8889",
+		/* USE_E2E_MOCKS=1 imports e2e/setup.ts, which mocks external services using MSW */
+		command: "CI=1 USE_E2E_MOCKS=1 pnpm dev --port 8889",
```
20jasper, line 77: Why not setup a dependent project to setup mocks instead of an env variable?

I can see this as valuable if we need mocks in dev locally

https://playwright.dev/docs/test-global-setup-teardown

20jasper (changes requested):

playwright.config.ts
fennifith: MSW needs to run in the same node process as the `fetch()` requests it needs to mock, so I'm not sure how that helps? ("setup" might be a misnomer here)

20jasper: Cool makes sense!

PR #1538, opened 2026-01-25
Fix e2e dependency errors

fennifith:
- Removes the `pnpm install --force` flag that causes pnpm to ignore the lockfile.
- Restructures the image/dockerfile to avoid needing the volumes & `user` property
- Adds a cache mount to `pnpm install` in the Dockerfile
- Filters out e2e deps from the prod build

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -353,7 +353,7 @@ importers:
   e2e:
     devDependencies:
       '@playwright/test':
-        specifier: ^1.58.0
+        specifier: 1.58.0
         version: 1.58.0
       cross-env:
         specifier: ^10.1.0


e2e/docker-compose.e2e.yml
```diff
@@ -35,10 +30,9 @@ services:
       dockerfile: e2e/Dockerfile
       target: e2e
     working_dir: /work/e2e
-    user: "${UID:-1000}:${GID:-1000}"
     volumes:
-      - .:/work/e2e:rw
-      - e2e_node_modules:/work/e2e/node_modules
+      - ./.playwright:/work/e2e/.playwright:rw
+      - ./tests:/work/e2e/tests:rw
```
20jasper, line 35: Does this need to be rw?

e2e/package.json
```diff
@@ -9,7 +9,7 @@
     "test:update-snapshots": "cross-env PLAYWRIGHT_ARGS=--update-snapshots pnpm test"
   },
   "devDependencies": {
-    "@playwright/test": "^1.57.0",
+    "@playwright/test": "^1.58.0",
```
20jasper: why not set this to an exact version?

PR #1540, opened 2026-01-26
TypeSense Search

crutchcorn:
This PR updates our search to use TypeSense due to Orama Cloud Legacy (what we were using) shutting down. This allows us to consolidate our deployment tooling.

I have verified that deploying to a local Docker instance sees search working nearly identically (exact results aside) from the current version.

# TODO

- [x] Update tests
- [x] Fix final failing tests
- [x] Handle Fly.io deployment
- [x] Handle proper ENV setting
- [x] Handle proper `read_only` token vs admin token
- [x] Handle proper `public_endpoint` setting
- [x] Remove Orama from current sponsors list

# Future TODOs

- [ ] Add back semantic search

### DIFF
--- a/.github/workflows/search-index.yml
+++ b/.github/workflows/search-index.yml
@@ -3,6 +3,7 @@ name: Deploy Search Index
 on:
   push:
     branches: [main]
+  workflow_dispatch:
 
 env:
   CI: true


build-scripts/search-index/deploy.ts
```diff
@@ -1,63 +1,103 @@
...
-	const isSnapshot = await index.snapshot(posts);
-	if (!isSnapshot) {
-		throw new Error("Unable to upload posts.");
+const existingCollections = await client.collections().retrieve();
+
+function findCollection(name: string) {
+	return existingCollections.find((c) => c.name === name);
+}
+
+function shallowCompare<T extends object>(a: T, b: T) {
```
20jasper, line 42: Is this correct? What if object a has 0 keys, and object b has 10? Would they not be the same here?

src/views/search/services.tsx
```diff
@@ -0,0 +1,227 @@
...
+		const facet = search.facet_counts?.find(
+			(facet) => facet.field_name === name,
+		);
+		if (!facet) return {};
+		return facet.counts.reduce(
+			(prev, count) => {
+				prev[count.value] = count.count + (prev[count.value] ?? 0);
+				return prev;
+			},
+			{} as Record<string, number>,
```
20jasper, line 201: Nit: Do you have to cast? Or could this be a type argument to reduce?

crutchcorn: I prefer the cast here. Not a fan of passing generics explicitly

PR #1547, opened 2026-02-03
Fix YouTube Embed in Previews

crutchcorn:
This PR:

- [x] Migrates us to use noembed.com to fix YouTube embed usage in prod and preview envs
- [x] Fixes issues with lazy images not loading in E2E tests
- [x] Makes 404 embed image lazy
- [x] Updates E2E tests

### DIFF
--- a/src/utils/markdown/data-providers/youtube.ts
+++ b/src/utils/markdown/data-providers/youtube.ts
@@ -45,11 +45,7 @@ export async function getYouTubeOEmbedDataFromUrl(
 		`https://noembed.com/embed?dataType=json&url=${encodeURIComponent(youtubeUrl)}`,
 	)
 		.then((res) => (res.ok ? res : Promise.reject(res)))
-		.then((res) => res.json())
-		.then((res) => ({
-			...res,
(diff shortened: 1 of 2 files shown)


src/utils/markdown/data-providers/youtube.ts
```diff
@@ -14,4 +14,42 @@ export interface YouTubeOEmbedResponse {
...
+	const youtubeUrl = `https://www.youtube.com/watch?v=${videoId}`;
+
+	return await fetch(
+		`https://noembed.com/embed?dataType=json&url=${encodeURIComponent(youtubeUrl)}`,
+	)
+		.then((res) => (res.ok ? res : Promise.reject(res)))
+		.then((res) => res.json())
+		.then((res) => ({
+			...res,
+			thumbnail_url: decodeURIComponent(res.thumbnail_url),
```
fennifith: What is this `decodeURIComponent` necessary for? Doesn't look like it should be needed unless there's something weird happening

(if there is something weird happening, I'd add a comment here pls)

crutchcorn: Good point. Lemme remove it from Vimeo and YouTube

PR #1554, opened 2026-02-05
Handle internal links for image wrapped links

crutchcorn:
This PR handles:

```md
<!-- ::start:link-preview -->
[Writing Modern JavaScript without a Bundler](/posts/modern-js-bundleless)
<!-- ::end:link-preview -->
```

By resolving our own domain properly.

# Future work

- [ ] Resolve post image from `getPostById` for the social image link

### DIFF
--- a/e2e/tests/posts-example.spec.ts-snapshots/posts-example-renders-dark-mode-1-firefox-linux.png
+++ b/e2e/tests/posts-example.spec.ts-snapshots/posts-example-renders-dark-mode-1-firefox-linux.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:a458d51c4017f310ec53a11b8700bdb8a319959c0a6ff6eb86b75f2976d3e0b0
-size 5237108
+oid sha256:1b247af2f594ec488b757a9ed72393d0deee3b77b5f1f4da01e4cb0b9f641e63
+size 5241484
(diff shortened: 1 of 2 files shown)


src/utils/markdown/components/link-preview/rehype-transform.ts
```diff
@@ -74,9 +75,14 @@ export const transformLinkPreview: RehypeFunctionComponent = async ({
...
-		? undefined
-		: (await getUrlMetadata(url.toString()).catch(() => undefined))?.banner;
+
+	const isPlayfulDomain = new URL(siteUrl).hostname === url.hostname;
+
+	const result = isPlayfulDomain
+		? { src: "/share-banner.png" }
+		: pictureNode
+			? undefined
+			: (await getUrlMetadata(url.toString()).catch(() => undefined))?.banner;
```
fennifith: Looks like this condition is the wrong way around - if `pictureNode` is available then we do not want to pass in a result (or it will show both images).

```suggestion
	const result = pictureNode
		? undefined
		: isPlayfulDomain
			? { src: "/share-banner.png" }
			: (await getUrlMetadata(url.toString()).catch(() => undefined))?.banner;
```

crutchcorn: Oh, yes, good call. Lemme update

PR #1578, opened 2026-02-27
[WIP] Mermaid diagram rendering support

### LINKED ISSUE #1298: Support MermaidJS embedding in articles
This issue is to support MermaidJS code embeds so that diagrams can be written in blog posts and rendered on the site:

https://mermaid.js.org/

For example, GitHub markdown supports these as a "```mermaid" code block: https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

fennifith:
Fixes #1298


native/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+const TERTIARY_COLOR: &str = "var(--opaque-surface_primary_emphasis-high)";
+
+const MERMAID_PIE_CHART_COLORS: [&str; 12] = [
+    PRIMARY_COLOR, SECONDARY_COLOR, TERTIARY_COLOR,
+    PRIMARY_COLOR, SECONDARY_COLOR, TERTIARY_COLOR,
+    PRIMARY_COLOR, SECONDARY_COLOR, TERTIARY_COLOR,
+    PRIMARY_COLOR, SECONDARY_COLOR, TERTIARY_COLOR,
+];
+
+lazy_static::lazy_static! {
```
20jasper: Consider `std::sync::OnceLock` isntead of lazy static. It's lighter, doesn't need a macro, plus it's in std!

native/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+rust-version = "1.93.1"
+
+[lib]
+crate-type = ["cdylib"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+neon = "1.1"
+mermaid-rs-renderer = { git = "https://github.com/1jehuang/mermaid-rs-renderer", tag = "v0.2.0", default-features = false }
```
20jasper: Huh seems weird that version isn't on cargo yet

native/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+    };
+
+    static ref RENDER_OPTIONS: RenderOptions = RenderOptions {
+        theme: THEME.clone(),
+        layout: LayoutConfig::default(),
+    };
+}
+
+#[neon::export]
+fn render_mermaid(document: String) -> String {
```
20jasper: Any reason not to use the JS interop types, eg
[JsResult](https://docs.rs/neon/latest/neon/result/type.JsResult.html). As far as I can tell, if the render fails now, it'll crash the process

PR #1616, opened 2026-06-25
upgrade to astrov7

20jasper:
Upgrade to astrov7, astro node v11 and astro preact v6. Draft since packages are too new to install now

Uses tsconfig paths from vitest directly

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved achievement loading on profile pages so achievement cards and the “view all achievements” option display more consistently.

* **Tests**
  * Refined search-page test coverage for sorting, pagination, filtering, rerendering, search reset, and collection behavior.
  * Updated serialization snapshots to reflect improved multiline formatting.

* **Chores**
  * Updated Astro, Node, and Preact integration versions.
  * Modernized TypeScript path resolution and test configuration.
  * Enabled stricter TypeScript module syntax checks and updated formatting exclusions.


src/utils/achievements/index.ts
```diff
@@ -8,190 +8,194 @@ export interface Achievement {
 	body: string;
 }
 
-export async function* getAchievements(
```
20jasper, line 11: For some reason no longer compiles. Generator is only used to iterate over to make a new array, so this removes one layer of indirection

src/utils/serialize-inline-call.node.spec.ts
```diff
@@ -18,8 +18,12 @@ test("serializes a function with arguments", () => {
 
 	expect(inline).toMatchInlineSnapshot(`
 		"((title, count, config) => {
-		  console.log("rust is a must 🦀🦀🦀🦀🦀");
-		  return { title, count, config };
+			console.log("rust is a must 🦀🦀🦀🦀🦀");
```
20jasper, line 21: I have no clue why this snapshot changed????

src/views/search/search-page.ui.spec.tsx
```diff
@@ -41,8 +41,11 @@ import {
 	PUBLIC_SEARCH_KEY,
 } from "./constants.ts";
 import Typesense from "typesense";
-import Collection from "typesense/lib/Typesense/Collection";
```
20jasper, line 44: These are no longer classes, not sure how this managed to pass before

fennifith (approved):

PR #1634, opened 2026-08-24
i18n Continued Work

crutchcorn:
image

> Translations above are AI generated and not included in this branch. Merely for testing

The following pages have had their UI localized:

- [x] Homepage
- [x] Person page
- [x] Events page
- [x] Join us page
- [x] Search page

However, the pages here that have outstanding UX concerns with localization (IE: User generated content) have been explicitly blocked on non-US routes

# TODOs

- [x] Migrate to [Paraglide](http://paraglidejs.com/astro)

# Future Work

- [ ] Locale switcher
- [ ] Warning in switcher that locale does not match `en.json`
- [ ] Fallback post/collection/person non-localized pages with `noindex` but allows user to see "Contents may not be translated for this page" and a localized non-usergen'd content?
- [ ] Snitips localization
- [ ] Person localization
- [ ] Allow searching by locale
- [ ] Limit homepage posts/collections by locale
    - Maybe some kind of "Only posts in X are shown. Search for more?" homepage text
- [ ] Custom collection pages


src/components/code-embed/code-embed.tsx
```diff
@@ -13,12 +13,14 @@ import type { TargetedEvent, ComponentChildren } from "preact";
...
 import type { FileEntry } from "./types.ts";
 import { ResizeablePanels } from "./resizeable-panels.tsx";
+import type { Translate } from "#utils/translations.ts";
 
 interface ContainerProps {
 	title?: string;
 	editUrl?: string;
 	codePanel: ComponentChildren;
 	previewPanel: ComponentChildren;
+	translate: Translate;
```
crutchcorn: This is the biggest question mark in my mind:

How do we want to handle localization for these TSX components that need to use dynamic `%s` interpolation?

src/utils/translations.ts
```diff
@@ -185,3 +230,7 @@ export function translate(
 
 	return value;
 }
+
+export function createTranslator(context: TranslationContext): Translate {
+	return (key, ...args) => translate(context, key, ...args);
+}
```
crutchcorn: Another open Q:

Will this pull in the entire translation JSON bundle size into JS for each page if it uses Preact?

I _think_ intuitively "yes", but how do we want to fix?

fennifith (approved):

PR #1658, opened 2026-08-27
lint(css): mark undeclared CSS variables

20jasper:
- inline styles with undeclared css vars
- stylelint to check for inline unknown variables, deprecated names, duplicate vars and more
  - I judiciuosly fixed some easy wins and left many be since they need work or could be itnentional
- screenshots for affected components

SCSS has modules, but CSS is global, so I just grab all the variables in all files and it works pretty good, but will have some false negatives prbably. This caught 30+ true positives from my brief validation pass, so it's working pretty well as is

Please see commit history for what real fixes were done. They are all low risk, but available to check!

## Quality Improvements

* Added automated visual checks for light and dark themes across key interface states.
* Added validation to detect unresolved CSS custom properties across application pages.
* Improved screen-reader-only styling with modern clipping behavior.

## Maintenance

* Improved stylesheet linting for SCSS, Astro, and CSS.
* Modernized equivalent CSS declarations and added a monospace fallback for code text.
* Refined layout spacing and decorative patterns without expected functional changes.


20jasper:
@coderabbitai review again thanks

PR #1660, opened 2026-08-28
chore: add git-lfs pre-push hook

20jasper:
I keep forget to push tha files, this will make it always push tha files


crutchcorn (approved):
Approved times a billion

PR #1661, opened 2026-08-28
chore: don't double minify CSS

20jasper:
Use lightning css bundled with vite instead of postcss + lightning css

- Build time: 26.5s to 16.1s (~39% faster, ~10.4s saved per build)

Sets browser list explicitly to widely available to avoid bundling legacy styles

- `document.css` (shared chunk): 72,942 B → 72,560 B minified (−382 B), 11,034 B → 10,920 B gzipped (−114 B)
- Total CSS across all 23 chunks: 505,283 B → 495,486 B minified (−9,797 B, −1.9%)
- Total CSS gzipped, all chunks: 93,069 B → 90,287 B (−2,782 B, ~−3%)

Validated this was due to removing some prefixes from pre chromium opera, Firefox, safari


crutchcorn (approved):
Double-checked that this didn't break FFG sites either; it doesn't. Merging

PR #1664, opened 2026-09-01
perf: load only required katex math fonts instead of fallbacks

20jasper:
Load only required math fonts instead of fallbacks. WOFF2 has been supported by major browsers for 9+ years (since Safari 10, 2016). Saves ~2KB per page with math. It also removes 800kb of unused fonts from the final docker container

AI Disclosre, I had Claude copy this from my personal site to here 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved math rendering in blog posts by ensuring KaTeX fonts load correctly across supported browsers.
  * Added comprehensive support for KaTeX font styles and weights.
  * Reduced font-related rendering issues by using bundled WOFF2 font files.


crutchcorn (approved):

20jasper:
I initially thought this saved more but browser is very kind and loads only what we need, but the inline styles are here, I guess I already did the work so we can keep it and take the 2kb savings

PR #1691, opened 2026-09-13
perf(code-embed): only preload initial file for code embeds

20jasper:
Migrate to fetching static files already in the server instead, pretender only initial file. Lowers initial page sizes and lowers build time by 10 seconds

Code embeds inlined full text in the HTML instead of querying. Web containers continue to query a zip file

The preview environment works fine, is it safe to assume production has the same nginx rules?

## uncompressed page sizes before interactions

The minimum bound is now lowered, if a user clicks every single file, it will be at most the same. Note that these compress well since they have lots of repeated files. The browser cache can now dedupe same files 

| page                                    | before  | after   |
|-----------------------------------------|---------|---------|
| ffg-fundamentals-side-effects           | 13.8 MB | 3.0 MB  |
| ffg-fundamentals-intro-to-components    | 11.6 MB | 2.5 MB |
| ffg-fundamentals-dependency-injection   | 10.0 MB | 2.1 MB  |
| post pages over 2 MB                    | 49      | 10      |

## benchmarks

Saves 10 seconds wall clock time. I believe this is faster due to IO and avoiding serializing props

| build         | wall | CPU (user+sys) | route generation |
|---------------|------|----------------|------------------|
| main          | 1:51 | 375 s          | 96 s             |
| this branch   | 1:41 | 367 s          | 87 s             |

## Summary by CodeRabbit

- **New Features**
  - Code embeds now load and display source files dynamically from the configured project location.
  - Selecting a file displays its filename while the corresponding content is loaded for viewing.
  - Small source files receive syntax highlighting when displayed.

- **Bug Fixes**
  - Improved code loading behavior to prevent outdated file content from appearing during selection changes.

- **Tests**
  - Added end-to-end coverage for viewing and switching between files in code embeds.

### DIFF
--- a/src/utils/markdown/components/code-embed/rehype-transform.ts
+++ b/src/utils/markdown/components/code-embed/rehype-transform.ts
@@ -82,13 +82,9 @@ export const transformCodeEmbed: RehypeFunctionComponent = async (props) => {
 	})) {
 		if (file.isFile()) {
 			const parent = path.relative(projectDir, file.parentPath);
-			const name = path.join(parent, file.name);
-			const code = await fs.readFile(path.join(projectDir, name), "utf-8");
-
 			files.push({
 				name: path.join(parent, file.name),
 				filetype: path.extname(file.name).substring(1),
-				code,
 			});
 		}
 	}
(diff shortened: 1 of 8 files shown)


20jasper:
https://pr-1691-playful-programming-preview-playfulprogramming.fly.dev/posts/ffg-fundamentals-intro-to-components/

It seems like embeds work fine. I don't perceive much of a delay on requests 

code embed with multiple tabs. The second tab is loaded and code is visible

src/utils/markdown/components/code-embed/rehype-transform.ts
```diff
@@ -106,13 +102,17 @@ export const transformCodeEmbed: RehypeFunctionComponent = async (props) => {
...
+				await fs.readFile(path.join(projectDir, fileEntry.name), "utf-8"),
+				fileEntry.filetype,
+			)
 		: undefined;
 
 	return [
 		createComponent("CodeEmbed", {
 			projectId: project,
 			projectZipUrl: `/generated/projects/${post}_${project}.zip`,
+			projectUrl: `/${path.relative(process.cwd(), projectDir)}`,
```
fennifith, line 115: I'm not sure I want to rely on these URLs always being accessible with the move to hoof - would be ideal if the component can use only `projectZipUrl`.

Would it be possible to build this into [webcontainer-script.ts](https://github.com/playfulprogramming/playfulprogramming/blob/42f8978f715388f9fa0b75e782b6b55ae058e2ab/src/utils/markdown/components/code-embed/webcontainer-script.ts)? Could use `i.fs.readFile` if the container is running, and otherwise fetch/cache the zip...

PR #1693, opened 2026-09-13
perf(api): parse each published date once when sorting posts

20jasper:
Centralizes post sorting. Originally this PR memorized 30M calls, but this moves it out of the front matter to avoid parsing per page load, so only 929 parse calls happen instead of n^2logn

Also makes post ordering deterministic (which is more valuable than the one second saved IMO)

Saves around a second

| | main | this branch |
|---|---|---|
| total wall | 87.4 s | 86.6 s |
| CPU (user+sys) | 233.8 s | 233.4 s |
| full-list sorts per build | ~2,000 | 1 per locale |

## Summary by CodeRabbit

* **Performance**
  * Improved caching for localized posts and collections, helping repeated filtering and sorting requests return more efficiently.
* **Consistency**
  * Standardized how localized content is read and cached across posts and collections.

### DIFF
--- a/src/utils/api.ts
+++ b/src/utils/api.ts
@@ -29,8 +29,22 @@ function findLocalizedEntry<T extends { locale: Locale }>(
 	);
 }
 
+// without cache, date objects are constructed thousands of times, taking several seconds
+const timestamps = new Map<string, number>();
+
+function toTimestamp(date: string): number {
+	const cached = timestamps.get(date);
+	if (cached !== undefined) {
+		return cached;
+	}
+
+	const timestamp = new Date(date).getTime();
+	timestamps.set(date, timestamp);
+	return timestamp;
+}
+
 function compareByDate(date1: string, date2: string): number {
-	return new Date(date1) > new Date(date2) ? -1 : 1;
+	return toTimestamp(date1) > toTimestamp(date2) ? -1 : 1;
 }
 
 function compareByPublished<T extends { published: string }>(


src/utils/api.ts
```diff
@@ -29,8 +29,22 @@ function findLocalizedEntry<T extends { locale: Locale }>(
...
+	}
+
+	const timestamp = new Date(date).getTime();
+	timestamps.set(date, timestamp);
+	return timestamp;
+}
+
 function compareByDate(date1: string, date2: string): number {
-	return new Date(date1) > new Date(date2) ? -1 : 1;
+	return toTimestamp(date1) > toTimestamp(date2) ? -1 : 1;
```
fennifith:
```suggestion
	return Date.parse(date1) > Date.parse(date2) ? -1 : 1;
```

Does this have a similar benefit if we just use `Date.parse` instead of maintaining our own cache?

20jasper: Parsing still dominates either way. I was surprised too. It's no faster in our specific use case (sorting 900ish date strings). I even tried 100k dates to sort, and it did not matter cached. I presume there's some some optimizations going on 

300 runs sorting 900 dates

| comparator                              | ms per sort |
|-----------------------------------------|-------------|
| `new Date(a) > new Date(b)` (before)    | 4.30        |
| `Date.parse(a) > Date.parse(b)`         | 2.29        |
| cached `Date.parse(...)`                | 0.26        |
| cached `new Date(...).getTime()` (this PR) | 0.24     |

20jasper: Too Long; Didn't Read--Date.parse *is* faster, but cache makes parsing a non factor

Okay dug in more with the profiler. Date parsing is superrr fast, we are just doing it a ton

Sorting builds new date objects per comparison, so nlogn calls with 2 calls per sort. With 929 posts to sort, that's 16,000ish. We have 929 places where the sort is done, so 30M parses done

It takes 300-500ns for a new Date().getTime() and 130-170ns with Date.parse. A map lookup is 15ns

So our limiting factor is what we do 30M times vs 929

20jasper: But that being said, we probably don't need to do anything 30 million times, we can go even deeper and avoid quadratic time complexity 

Thanks for pushing back on this one! This is def a bandaid fix

src/utils/api.ts
```diff
@@ -29,15 +29,46 @@ function findLocalizedEntry<T extends { locale: Locale }>(
...
+	if (result === undefined) {
+		result = Promise.all(
+			[...collections.values()]
+				.map((locales) => findLocalizedEntry(locales, language))
+				.filter(isDefined)
+				.map((c) => readCollection(c)),
+		).then((all) => all.sort(compareByPublished));
+		collectionsByLocale.set(language, result);
+	}
+	return result;
```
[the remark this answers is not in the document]
20jasper: We specifically wanted to avoid caching there for complexity reasons, and anyways, this will save on the order of milliseconds since parsing is on the order of nanoseconds and we are only comparing once per locale now

PR #1700, opened 2026-09-13
perf(build): compress project zips natively

20jasper:
`fflate.zip()` only offloads files over 160 KB to a worker, meaning every file is deflated in JS on the main thread

Files are now compressed natively, in parallel, on the libuv threadpool with `zlib.deflateRaw`

Also lowers compression level from 9 to 6, which is 2.5x faster than the js version, *and* compresses better than the js version. We can lower compression level further if we want to save more time

Saves 5 seconds wall clock but 45 seconds CPU time. Avoids unnecessary GC by not using the V8 heap

Measures compressing every file in the project

| compressor         | level | output  | time   |
|--------------------|-------|---------|--------|
| fflate (JS) main  | 9     | 58.2 MB | 4.53 s |
| fflate (JS)        | 6     | 58.3 MB | 4.12 s |
| fflate (JS)        | 4     | 59.7 MB | 3.40 s |
| zlib (native)      | 9     | 55.8 MB | 4.02 s |
| zlib (native) this branch | 6     | 56.4 MB | 2.10 s |
| zlib (native)      | 4     | 58.4 MB | 1.53 s |
| zlib (native)      | 1     | 62.1 MB | 1.10 s |

EDIT: this pr also bumps to node 26, which has perf benefits across the board

Node 24.16 → 26.8.2 alone (main @ 48d2ad73e

| | Node 24.16 | Node 26.8.2 | Δ |
|---|---|---|---|
| wall | 92.5 s / 92.5 s | 85.6 s / 85.2 s | −7.1 s (−8 %) |
| CPU (user+sys) | 248 s / 251 s | 232 s / 229 s | −19 s (−8 %) |

Pipeline bench, single thread, all 912 posts 

| pass | Node 24 | Node 26 | Δ |
|---|---|---|---|
| total | 58.8 s | 54.4 s | −7 % |
| rehypeShikiUU | 33.5 s | 31.4 s | −6 % |
| rehypeTransformComponents | 9.5 s | 8.5 s | −11 % |
| parse | 6.0 s | 5.2 s | −13 % |
| rehypeAstroImageMd | 0.78 s | 0.67 s | −14 % |
| remaining passes | | | −1 … −7 % |

## Summary by CodeRabbit

* **New Features**
  * Added support for creating reproducible ZIP archives from in-memory files.

* **Compatibility**
  * Added support for Node.js 26.8 ZIP compression APIs.
  * Updated runtime requirements to Node.js 26.8.x.

* **Chores**
  * Updated development and end-to-end container environments to Node.js 26.8.
  * Improved Corepack setup across containerized environments.

### DIFF
--- a/src/pages/generated/projects/[slug].zip.ts
+++ b/src/pages/generated/projects/[slug].zip.ts
@@ -3,7 +3,9 @@ import { getAllPosts, getPostBySlug } from "#utils/api.ts";
 import path from "path";
 import { contentDirectory } from "#utils/data.ts";
 import fs from "fs/promises";
-import { zip } from "fflate";
+import { Zip, type ZipInputFile } from "fflate";
+import { crc32, deflateRaw } from "node:zlib";
+import { promisify } from "node:util";
 import { baseLocale } from "#src/paraglide/runtime.js";
 
 export async function findProjectDir(slug: string): Promise<string> {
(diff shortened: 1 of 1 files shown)


src/pages/generated/projects/[slug].zip.ts
```diff
@@ -3,7 +3,8 @@ import { getAllPosts, getPostBySlug } from "#utils/api.ts";
 import path from "path";
 import { contentDirectory } from "#utils/data.ts";
 import fs from "fs/promises";
-import { zip } from "fflate";
+import { ZipEntry, createZipArchive } from "node:zlib";
```
Comment (20jasper) [inline src/pages/generated/projects/[slug].zip.ts]: We need node 26.8 for createZipArchive

Dockerfile
```diff
@@ -1,12 +1,12 @@
...
-FROM node:24-alpine3.24 AS builder
+FROM node:26.8-alpine3.24 AS builder
 
 # Create app directory
 WORKDIR /var/app
 
 # Prepare pnpm according to the root package.json
 COPY package.json .
-RUN corepack enable
+RUN npm install --global corepack && corepack enable
```
20jasper, line 9: Core pack is not bundled with node anymore in node 26

fennifith (approved):

PR #1710, opened 2026-09-19
Use VFile Failures for Logging

crutchcorn:
This PR introduces consistent logging that's not edgecased against our own plugins. This means that `vfile` logging is more consistent even with external Unified plugins.

Open for feedback / pushback, but this PR makes more sense in broader contexts of MDAST Extension work

Disclaimer: This code was written using AI, but every LOC was reviewed by myself and the direction was set by myself and took a lot of "offline" (IE: not shown in this history) wrangling of the Astra 6 Ultra model.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Markdown validation so issues are reported consistently with clearer locations and rule details.
  - Preserved validation behavior for invalid links, images, embeds, components, quizzes, diagrams, and frontmatter.
  - Invalid content is now handled more gracefully without obscuring other processing errors.

- **Developer Experience**
  - Linting results now provide more reliable warnings across people, collections, and posts.
  - Diagnostics include relevant source context and are better integrated with continuous-integration reporting.


src/utils/markdown/diagnostics.ts
```diff
@@ -0,0 +1,112 @@
...
+	return value
+		.replaceAll("%", "%25")
+		.replaceAll("\r", "%0D")
+		.replaceAll("\n", "%0A");
+}
+
+function reportMessages(file: VFile, contents: string) {
+	for (const message of file.messages) {
+		if (reportedMessages.has(message)) continue;
+		reportedMessages.add(message);
```
fennifith, line 43: Can `reportedMessages` be removed? This will cause issues keeping global state in SSR. I don't see any problem with logging a message twice if it's been reported twice.

src/utils/markdown/lintMarkdown.ts
```diff
@@ -0,0 +1,25 @@
...
+) {
+	const file = await getMarkdownVFile(stub);
+	await withMarkdownDiagnostics(file, async () => {
+		try {
+			const post = await read(stub, file);
+			await getMarkdownHtml(post, file);
+		} catch (error) {
+			// A fatal diagnostic is a lint result; unrelated failures still propagate.
+			if (
+				!file.messages.some((message) => message === error && message.fatal)
```
fennifith, line 18: Could this be `if (error instanceof VFileMessage)` (or some `isVFileMessage` util) to make this clearer?

src/utils/markdown/getMarkdownHtml.ts
```diff
@@ -17,10 +18,12 @@ export async function getMarkdownHtml(
 ): Promise<MarkdownHtml> {
 	const vfile = await vfilePromise;
 
-	const result = await unifiedChain.process(vfile);
-
-	return {
-		...vfile.data,
-		content: (await result.result) as components.PlayfulNode[],
-	};
+	vfile.data.frontmatter = post;
```
fennifith, line 21: Why are we setting `vfile.data.frontmatter` here? Can this be removed?

src/utils/markdown/diagnostics.ts
```diff
@@ -0,0 +1,112 @@
...
+				kleur.gray(
+					`\tin ${filePath || "<markdown>"}${message.line === undefined ? "" : `:${message.line}`}`,
+				),
+			);
+		}
+	}
+}
+
+/** Report every plugin's messages, including when processing stops early. */
+export async function withMarkdownDiagnostics<T>(
```
fennifith, line 102: More of a nit - I'm iffy about this being a wrapper as it obscures the internals even further (I think the split between `getMarkdownVFile -> readPost -> unifiedChain.process` is already confusing)

I'd generally prefer this as an imperative util, i.e. transform usages to:

```js
const reporter = createReporter(vfile);
const result = await unifiedChain.process(vfile)
    .then(reporter.success)
    .catch(reporter.failure);
```

This way the control flow from read -> process is directly shown in one place.

src/utils/markdown/lintMarkdown.ts
```diff
@@ -0,0 +1,25 @@
...
+import { getMarkdownHtml } from "./getMarkdownHtml.ts";
+import { getMarkdownWarnings, withMarkdownDiagnostics } from "./diagnostics.ts";
+import type { MarkdownFileInfo, MarkdownVFile } from "./types.ts";
+
+export async function lintMarkdown<Stub extends MarkdownFileInfo>(
+	stub: Stub,
+	read: (stub: Stub, file: MarkdownVFile) => Promise<MarkdownFileInfo>,
+) {
+	const file = await getMarkdownVFile(stub);
+	await withMarkdownDiagnostics(file, async () => {
```
fennifith, line 11: Potentially out of scope here - IMO `lintMarkdown` shouldn't actually log/report the errors itself, that's kind of an obscured side-effect at the moment. I think it makes more sense to just return the errors as-is, and move the logging into `scripts/lint-markdown-files.ts` where it processes the response.

src/utils/markdown/diagnostics.ts
```diff
@@ -0,0 +1,112 @@
...
+			message: message.reason,
+			path: messagePath(file, message),
+			offset:
+				message.place && "start" in message.place
+					? message.place.start.offset
+					: message.place?.offset,
+			col: message.column,
+			line: message.line,
+		}));
+}
```
fennifith, line 31: Another small note - the lint script in CI is the only consumer of the lint API. It's much more preferable to just change the response shape and handle it in the script rather than adding an unnecessary filter/map.

(also, why is this filtered to only fatal messages? warnings should include all messages, no?)

PR #1711, opened 2026-09-19
Move demo html into iframes

crutchcorn:
This PR addresses some of the errant tags in https://github.com/playfulprogramming/playfulprogramming/issues/1689 by moving inline HTML demos to inline iframes.

There's a minor A11Y hit that occurs thanks to `frame` being read out, but it doesn't seem too bad from my testing, and cannot be worked around.

## Summary by CodeRabbit

* **New Features**
  * Added inline, borderless iframe previews that fill the article width and automatically fit their content height.
  * Added numerous interactive accessibility examples covering labels, forms, buttons, semantic markup, text sizing, contrast, and readability.
  * Added styling demonstrations for gradients, readable colors, image borders, and image blur effects.
* **Documentation**
  * Documented how to enable inline iframe previews with the `data-no-frame` attribute.


content/crutchcorn/collections/art-of-accessibility/posts/art-of-a11y-labels/index.md
```diff
@@ -43,11 +43,7 @@ By default, this will look like the following:
 ----
 
 
-<form>
-    <input name="username" type="text"/>
-    <input name="password" type="password"/>
-    <button type="submit">Login</button>
-</form>
+<iframe src="/custom-content/art-of-accessibility/art-of-a11y-labels/unlabeled-form.html" data-no-frame></iframe>
```
fennifith, line 46: Can these files all be moved into the post folder? i.e. this src should be `"./unlabeled-form.html"`, directly adjacent to the current post.

Hoof will not pick up attachments if they aren't adjacent/within the post dir, as it needs to upload/index the content before the full markdown is parsed.

These paths probably need resolving with `resolvePath` the same way we do for images.

src/utils/markdown/iframes/rehype-transform.ts
fennifith, line 59: Another note - `srcReplacements` is currently empty in every configuration. Can we just remove this?

